PR #9, opened 2017-07-16
need python3.x

hadfl:


postwait:
I am unclear on what omnios-extra is. The goals originally set out for omnios were to be minimal. The hope was to. Remove perl, python and Java completely. Obviously pkg(5) make python 2 removal hard, but I see a lot of stuff being added. Might we worth reviewing the guiding principles or revising them.

hadfl:
The intention for omnios-extra is to contain packages we need to run our `omniosce.org` infrastructure. It is not part of OmniOS so does not violate the approach to be minimal.

At the moment, we also add packages that are not in core but necessary to build OmniOS.
I totally agree with you that we should rather eliminate the dependencies than adding packages here but for now, to be able to move ahead this is the least critical approach.

We leave the issues open so we will look at it once everything settled a bit and we have spare cycles to tackle these: https://github.com/omniosorg/omnios-build/issues/7 https://github.com/omniosorg/omnios-build/issues/6

danmcd:
This bad-case is when two packages upon which you depend (pkg5 and iso-codes) rely on two entirely-incompatible versions of a piece of software (Python 2.7 and 3.x in this case).

Also, Python3 is a built-time dependency, not a run-time dependency.  OpenJDK has this problem also with ANT.

citrus-it:
Dan I think you and @hadfl are agreeing. We know that we need python3, ant and freetype2 to build Omnios so these packages are going in extra, a completely separate IPS repo. There won't be too much more in there although if and when we decide to remove tcsh, zsh, tmux, screen, java, perl, ... or whatever else, we might move those to extra at least in the short term to ease the transition.

.gitignore
```diff
@@ -2,3 +2,4 @@ build.log
 build.log.1
 tmp.repo/
 *.p5m.final
+build/python36/local.mog
```
citrus-it: Why not add this to a .gitignore file inside build/python36 rather than at the top level?

hadfl: done.

PR #11, opened 2017-07-17
Add mediators to python links

citrus-it:
This change makes the symlinks to python in `/opt/ooce/bin/` and `/opt/ooce/share/man/` into mediated links, allowing multiple omnios-extra versions of python to be installed in parallel, should the need arise.

```
bloody# pkg mediator -a
MEDIATOR VER. SRC. VERSION IMPL. SRC. IMPLEMENTATION
gcc      system    5.1     system
mta      system            system     mailwrapper
mta      system            system     sendmail
python   system    3.6.1   system
```

Perhaps we should call the mediator ooce-python?


oetiker:
in the case of python I would like to see multiple symlinks

python
python3
python36
python2
python27

each with mediaton then your script could run on /opt/ooce/bin/python3

hadfl:
replaced by https://github.com/omniosorg/omnios-extra/pull/12

PR #17, opened 2017-09-10
packages info added

hadfl:
I've started a `doc` branch. Let's not merge this until it includes all necessary info so users can start contribute/maintain packages.

Please push your documentation (commits) to this branch. We can squash it when merging.

### DIFF
--- a/doc/structure.md
+++ b/doc/structure.md
@@ -1,0 +1,47 @@
+# About Packages on OmniOSce Extra
+
+The extra repository contains prime quality packages to enhance your omnios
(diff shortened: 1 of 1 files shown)


oetiker:
I have added a structure.md with some ideas ... regarding extra repo structure. what do you think about the multiple version thing ? It would also make it very easy to add new stuff without fear of breaking existing things ... people would have to manually upgrade individual packages if they decided that they wanted to go to a newer version.

citrus-it:
@hadfl - can we get this merged so we can point potential contributors at the documentation?

PR #46, opened 2017-11-01
fix github downloading release URL

seriv:
It can be verified by curl. 
<code>curl -D - https://github.com/hadfl/ooceapps/archive/v0.1.4.tar.gz</code> works fine (gives 302 redirect), while <code>curl -D - https://github.com/hadfl/ooceapps/releases/download/v0.1.4.tar.gz</code> gives 404.

### DIFF
--- a/build/ooceapps/build.sh
+++ b/build/ooceapps/build.sh
@@ -36,7 +36,7 @@ SUMMARY="Mattermost integrations for OmniOSce" # One-liner, must be filled in
 DESC=$SUMMARY   # Longer description, must be filled in
 BUILDARCH=64    # or 64 or both ... for libraries we want both for tools 32 bit only
 PREFIX=/opt/ooce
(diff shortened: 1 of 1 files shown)


hadfl:
Thanks for reporting this issue. The reason that it fails to download is because we are in the process of moving `ooceapps` from my GitHub account to `omniosorg` (I wiped the releases from my own repository already).

The proper fix would be `s/hadfl/omniosorg/` in the URL, could you please adjust your fix accordingly?
You might as well update the version to `0.2.0` as this is the first version we drafted a release for on `omniosorg`.

On a side note: we are downloading the bundled package which contains all the perl modules which is e.g. `ooceapps-0.2.0.tar.gz`, not `v0.2.0.tar.gz` those are only present in the releases subdirectory.

seriv:
Ok, I will recreate push request.

PR #47, opened 2017-11-01
fix github URL

seriv:
Ooceapps repository moved to omniosorg and releases are wiped out in the original location. Also, version changed to be 0.2.0 which is the first release made for omniosorg.


build/ooceapps/build.sh
```diff
@@ -54,7 +54,7 @@ add_extra_files() {
 }
 
 init
-download_source "v$VER" $PROG $VER
+download_source "v$VER" $PROG $PROG
```
hadfl: this should not be changed.

hadfl: https://github.com/omniosorg/omnios-extra/pull/47/files#diff-2df8302a6d21884fc141f74199cfce28R31 << but version should be changed here to `VER=0.2.0`

hadfl:
closing this as there was no feedback. fixed in https://github.com/omniosorg/omnios-extra/pull/48

PR #89, opened 2018-01-10
drop 'etc/opt' globally (r151022)

hadfl:


lib/global-transforms.mog
```diff
@@ -86,6 +86,7 @@
 <transform dir path=etc$ -> drop>
 <transform dir path=etc/opt$ -> drop>
 <transform dir path=etc/inet$ -> set group sys>
+<transform dir path=etc/opt$ -> drop>
```
oetiker: in what way is this line different than the one 2 lines above ?

hadfl: you are right. it was only missing in `master` as it got deleted when taking in build env enhancements from `omnios-build`. created PRs to revert fro r22/r24

PR #103, opened 2018-01-17
adding mbuffer

hadfl:


build/mbuffer/patches/series
```diff
@@ -0,0 +1 @@
+no-gcrypt.patch
```
oetiker: in order to activate the new configure you have to call `autoconf`

hadfl: fixed.

PR #116, opened 2018-02-06
 smartmontools added (r151022)

hadfl:


build/smartmontools/files/system-smartd.xml
```diff
@@ -0,0 +1,71 @@
...
+        <service_fmri value="svc:/system/filesystem/autofs:default"/>
+    </dependency>
+    <dependency name="config-file" grouping="require_all" restart_on="refresh" type="path">
+        <service_fmri value="file://localhost/etc/opt/ooce/smartmontools/smartd.conf" />
+    </dependency>
+
+    <exec_method
+        type="method"
+        name="start"
+        exec="/opt/ooce/sbin/smartd --pidfile=/var/opt/ooce/smartmontools/run/smartd.pid --savestates=/var/opt/ooce/smartmontools/"
```
oetiker: there is a /var/run directory, mounted to swap, so that it gets flushed on reboot ... I guess it would make sense to have our pid files there too ... `/var/run/opt/ooce/...`  as this avoids stale pid files after reboot.

hadfl: well we'd need to add wrapper scripts to create the subdirectories in `/var/run` if they don't exist (i.e. after every reboot).

citrus-it:
On Tue, 6 Feb 2018, Dominik Hassler wrote:

; hadfl commented on this pull request.
;
;
;
; > +    <single_instance/>
; +
; +    <dependency name="filesystem-local" grouping="require_all" restart_on="none" type="service">
; +        <service_fmri value="svc:/system/filesystem/local:default"/>
; +    </dependency>
; +    <dependency name="autofs" grouping="optional_all" restart_on="error" type="service">
; +        <service_fmri value="svc:/system/filesystem/autofs:default"/>
; +    </dependency>
; +    <dependency name="config-file" grouping="require_all" restart_on="refresh" type="path">
; +        <service_fmri value="file://localhost/etc/opt/ooce/smartmontools/smartd.conf" />
; +    </dependency>
; +
; +    <exec_method
; +        type="method"
; +        name="start"
; +        exec="/opt/ooce/sbin/smartd --pidfile=/var/opt/ooce/smartmontools/run/smartd.pid --savestates=/var/opt/ooce/smartmontools/"
;
; well we'd need to add wrapper scripts to create the subdirectories in `/var/run` if they don't exist (i.e. after every reboot).

And then this only works for things running as root since other processes
won't be able to create the necessary directories.
I would stick with /var/opt/ooce/<prog>/run/ or, if the software does not
properly handle stale pid files, fix it or place the pid file in /tmp

-- 
Citrus IT Limited | +44 (0)333 0124 007 | enquiries@citrus-it.co.uk
Rock House Farm | Green Moor | Wortley | Sheffield | S35 7DQ
Registered in England and Wales | Company number 4899123

oetiker:
@citrus-it just wondering why linux and solaris (at least) have introduced tmpfs backed '/var/run' in the last few years. for log files we go for /var/log/opt ... so ... one option would be to have a ooce service that takes care of setting up such directories for all ooce packages that want them ... (configurable via properties in the respective smf manifest of the package) ...

citrus-it:
Well, from `man filesystem`:
>       /var/run
>
>           Temporary files which are not needed across reboots. Only root may
>           modify the contents of this directory.

It isn't the place for our daemons to be putting stuff IMHO - and we should always be aiming to start things as non-root and with even further reduced privs if possible.

oetiker:
ok ... lets leave it in non temporary space then

PR #138, opened 2018-03-02
docbook: add docbook xsl package

citrus-it:
This provides local copies of stylesheets that can be used when generating man pages using asciidoc instead of retrieving them from sourceforge.net

### DIFF
--- a/build/docbook/build.sh
+++ b/build/docbook/build.sh
@@ -16,38 +16,36 @@
 
 . ../../lib/functions.sh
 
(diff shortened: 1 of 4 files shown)


tomww:
I would love to see this package in regular OmniOSce repository.
Rationale: If not added to the main OS IPS repository, then every 3rd-party software stack using docbook would have to add the omniosce-extra IPS repository only for this package.
Thank you!

PR #148, opened 2018-04-11
fix path to postfix

moetiker:
postfix is in  /opt/ooce/postfix  located

### DIFF
--- a/build/postfix/files/smtp-postfix.xml
+++ b/build/postfix/files/smtp-postfix.xml
@@ -18,19 +18,19 @@
         <exec_method
             type="method"
             name="start"
-            exec="/opt/ooce/sbin/postfix start"
+            exec="/opt/ooce/postfix/sbin/postfix start"
             timeout_seconds="60">
             <method_context security_flags="aslr" />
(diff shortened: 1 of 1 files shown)


oetiker:
two things come to mind ... 

* is there a reason for the service not starting (or even worse some other postfix starting) if the link is not set ?

* how does a user of this package know about the existence of this setting and the requirement to activate it prior to enabling postscript?

hadfl:
yes there is a reason to prevent the service from being started if "helper" programs like `mailq` and `sendmail` are not "linked" to that service but a different MTA like dma or sendmail. about starting "another" postfix: that part is a bit trickier. probably requiring to mediate SMF manifests. not sure how to solve that one. but `/opt/ooce` is in our "namespace" so we should be safe.

the MTA in omnios has been mediated since `mailwrapper` was added in 2014. so this should not be big news and if we provide additional MTAs in the -extra repo i think those should be mediated likewise.

moetiker:
thanks for the information about 'pkg set-mediator -I postfix mta' to create the sbin links.

PR #157, opened 2018-05-23
Nodejs v8.11.2

cneira:
Added Node v8.11.2 to Omnios Extras from https://nodejs.org/dist/v8.11.2/node-v8.11.2.tar.gz

### DIFF
--- a/build/nodejs/build.sh
+++ b/build/nodejs/build.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+
+# Copyright 2018 OmniOS Community Edition (OmniOSce) Association.
+# Use is subject to license terms.
+#
+# Load support functions
+. ../../lib/functions.sh
+
+PROG=node
+VER=8.11.2
+VERHUMAN=$VER
+PKG=ooce/runtime/node-811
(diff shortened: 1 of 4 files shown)


build/nodejs/build.sh
```diff
@@ -0,0 +1,62 @@
...
+VERHUMAN=$VER   # Human-readable version
+#PVER=          # Branch (set in config.sh, override here if needed)
+PKG=developer/nodejs            # Package name (e.g. library/foo)
+SUMMARY="Node.js is an evented I/O framework for the V8 JavaScript engine." # One-liner, must be filled in
+DESC="Node.js is an evented I/O framework for the V8 JavaScript engine.
+It is intended for writing scalable network programs such as web servers. " # Longer description, must be filled in
+
+BUILDARCH=64
+OPREFIX=$PREFIX
+PREFIX+="/$PROG-$VER"
```
oetiker: I would suggest to use a 2 digit version ... so that it ends up in 

`PREFIX+=/${PROG}-8.11`

the patch releases can then be treated as updates to this package while 8.12 would be a new package ...

citrus-it: nginx is a good template for this, it includes:

```
MAJVER=${VER%.*}            # M.m
sMAJVER=${MAJVER//./}       # Mm
PREFIX+=/$PROG-$MAJVER
```

build/nodejs/build.sh
```diff
@@ -0,0 +1,61 @@
...
+CXXFLAGS+="-ffunction-sections -fdata-sections"
+MAKE_ARGS="CC=$CC"
+
+CONFIGURE_OPTS_64=" \
+    --with-dtrace \
+    --dest-cpu=x64 \
+    --prefix=$PREFIX \
+"
+BUILDDIR=$PROG-v$VER
+DEPENDS_IPS="system/library system/library/gcc-runtime   developer/gnu-binutils"
```
citrus-it: The first two dependencies are automatically detected so no need to specify them by hand.
Is gnu-binutils a runtime dependency (I know it's a build-time one).
If it's just for build, specify it as `BUILD_DEPENDS_IPS="developer/gnu-binutils"`

PR #174, opened 2018-09-26
libpciaccess: PCI access utility library

drscream:
The pciaccess library wraps platform-dependent PCI access methods in a convenient library. For exmaple it's required to build gfx-drm and the kernel module.

### DIFF
--- a/build/libpciaccess/build.sh
+++ b/build/libpciaccess/build.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+#
+# Copyright 2018 Thomas Merkel
+#
+# Load support functions
+. ../../lib/functions.sh
+
+PROG=libpciaccess
+VER=0.14
+VERHUMAN=$VER
+PKG=library/libpciaccess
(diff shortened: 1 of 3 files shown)


oetiker:
make sure you take note of https://github.com/omniosorg/omnios-extra/blob/master/doc/structure.md

drscream:
@oetiker thanks a lot for the hint, I will have a look and update my pull request as recommended. Is the version number suffix mandatory, because it's not used by lot's of other package in the build folder?

oetiker:
Whether a package uses a version number depends on the package and its evolution ... if your package contains a shared library, the naming choice of both the package and the directory will determine your path for upgrading it when new versions come out. If it is a library where multiple versions could conceivably co-exist on the same system, you should take care that this is possible ... 

by incorporating the major version of the library into the package name for example (and the directoy too) ... 

in other words, there is no hard and fast rule but you are required to decide what works best for your package

drscream:
I've changed the `PKG` name based on your request. I hope the following installation is okey for the package:

```
PATH
opt/ooce
opt/ooce/include
opt/ooce/include/pciaccess.h
opt/ooce/lib
opt/ooce/lib/amd64
opt/ooce/lib/amd64/libpciaccess.a
opt/ooce/lib/amd64/libpciaccess.so
opt/ooce/lib/amd64/libpciaccess.so.0
opt/ooce/lib/amd64/libpciaccess.so.0.11.1
opt/ooce/lib/amd64/pkgconfig
opt/ooce/lib/amd64/pkgconfig/pciaccess.pc
opt/ooce/lib/libpciaccess.a
opt/ooce/lib/libpciaccess.so
opt/ooce/lib/libpciaccess.so.0
opt/ooce/lib/libpciaccess.so.0.11.1
opt/ooce/lib/pkgconfig
opt/ooce/lib/pkgconfig/pciaccess.pc
```

drscream:
@hadfl I've fixed the order, sorry for the delay.

PR #194, opened 2018-12-07
Adding rust and ripgrep

citrus-it:


citrus-it:
Well, that was fun... the only change is actually that I have fixed the man page location.

PR #227, opened 2019-01-20
ripgrep is now required to build omnios userland

citrus-it:


citrus-it:
Not required, this was fixed in #229

PR #274, opened 2019-03-15
virtualbox: add exclusive hma registration support

citrus-it:


oetiker:
is this connected to preventing virtual box to run while bhyve is the law of the land?

citrus-it:
> is this connected to preventing virtual box from running while bhyve is the law of the land?

Yes. At the moment it is possible to start a Virtualbox VM while there are bhyve/kvm ones running, or vice versa. So it's up to the administrator to stop that somehow, or we have to ship /etc/system fragments with virtualbox that exclude the bhyve and kvm modules loading. That is not great as it prevent easy switching between systems.

This (and the corresponding kernel PR which is in progress) change this so that the system enforces the rules. bhyve or KVM can be started as long as all vbox VMs are stopped, and no vbox VM can start up if any bhyve/kvm VM is running.

PR #294, opened 2019-04-12
Add rust 'fd' and 'bat' utilities.

citrus-it:
With many thanks to @papertigers for the assistance in building these, and for the work he is doing to add illumos support to rust crates.

### DIFF
--- a/build/bat/build.sh
+++ b/build/bat/build.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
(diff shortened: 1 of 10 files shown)


priyadarshan:
Thank you so much for these, they have become a staple in my toolchain, thanks to @papertigers  too. Looking forward to see the [nix crate](https://github.com/nix-rust/nix/issues/935) available on illumos!

papertigers:
You *should* be able to build bat with just the termios change. I think the libc stuff has all made it into the libc crate itself via @jasonbking. See https://github.com/rust-lang/libc/pull/1251

citrus-it:
> You _should_ be able to build bat with just the termios change. I think the libc stuff has all made it into the libc crate itself via @jasonbking. See [rust-lang/libc#1251](https://github.com/rust-lang/libc/pull/1251)

Without your branch, I hit 
```
39 |         libc::ioctl(libc::STDOUT_FILENO, libc::TIOCGWINSZ.into(), &mut winsize);
   |                                                ^^^^^^^^^^ not found in `libc`
```
but it looks like a `cargo update libc` was sufficient, thanks

PR #472, opened 2020-04-09
add tree-1.8.0 build files

pbdigital-dot-org:
Tree is a recursive directory listing command that produces a depth indented listing of files, which is colorized ala dircolors if the LS_COLORS environment variable is set and output is to tty. 
Tree has been ported and reported to work under the following operating systems: Linux, FreeBSD, OS X, Solaris, HP/UX, Cygwin, HP Nonstop and OS/2.  And if all this looks good, now for OmniOSce.

### DIFF
--- a/build/tree/build.sh
+++ b/build/tree/build.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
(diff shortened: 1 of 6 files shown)


pbdigital-dot-org:
HI @hadfl ,
Thanks for your kind feedback. I have made those changes and pushed them up to b9faab2.
Philip

PR #477, opened 2020-04-17
add fcgiwrap-1.1.0 build files

pbdigital-dot-org:
add library/fcgi2 & application/fcgiwrap for nginx/nagios

I have Nagios running and packaged (with nagios-plugins,nrdp & nsca), which I would like to push up to omnios-extra.
Nagios is running under nginx and requires fcgiwrap, so I have made PRs for the fcgi parts today. Any pointers I get on these PRs 
I can apply to the Nagios packages so as not to consume too much of your time.

for sure, one thing I was not certain on was UID & PID for user fcgiwrap. 

Philip


citrus-it:
Thanks a lot of raising this PR. I took a look at it and had a few comments, some fixes and some suggestions. To make it easier to review I've created a branch and added this commit to it:

https://github.com/citrus-it/omnios-extra/commit/c1ef4423c8aa8bb6f2144d5da72ff7fddd1a5d64

Please feel free to take or leave the suggestions - here are my notes on the changes:

1. Although it's tempting to use $PROG in a build.sh's PKG= line, external
  scripts extract these PKG lines (including travis, which is why this PR is showing with failed checks).

2. The licence here seems to be a standard MIT licence (you can always run
   `../../tools/licence path/to/licence` to see if the system identifies it).
   The main point of the licence check is to catch packages which change their
   licence with a new version to make sure we get this correctly reflected in
   the package metadata.
   For this reason, it's better to extract the licence from the source if
   at all possible. I've removed the licence file from files/ and added a
   function to build.sh to do this for you. Then adjusted the local.mog and
   the licence type and finally removed the SKIP_LICENCES line from build.sh
  since it's a known licence type.

3. Where a dependency is auto-detected, it's better to let this happen than
   to hard code it - I've removed RUN_DEPENDS_IPS. I also changed the
   BUILD_DEPENDS_IPS to += rather than overwriting whatever might be there.

4. I changed to CONFIGURE_OPTS which is where arch-independent options should
   go. The makefile that comes with this package is broken in terms of what
   it expects --sbindir to look like.

5. The fcgi2 package should be delivered as multi-arch
   (we do this with libraries) so I've changed the LDFLAGS line to use the
   amd64/ component. @hadfl is going to review the fcgi2 PR this evening.

6. Downloading the source from github creates an archive with just the version
   number in the name, so I've changed the download_source line (and also put
   the archive on the omnios mirror)

7. I dropped the service method script and did it all in the SMF manifest.
   While I was on, I added a service property to let administrators change
   the number of children (-c argument).
   (`startd/duration=child` is what tells it that the process will not
   background itself.)

8. I noticed that the patches were applying with fuzz so I did a re-base
   (./build.sh -P) - this also sets the dates to 1970 so that future updates
    and re-basing don't change too much.

9. The control socket is not cleaned up when the process exits, and it's on
   non-tmpfs. This causes two problems:
   - if the service somehow dies or the system crashes (i.e. the method stop
     is not run) then the service will not come back until somebody manually
     removes the file.
   - if the boot environment is cloned for an upgrade, then when booting to the
     new BE, the socket will already exist and the service will fail.

   To fix these, I added a patch to fcgiwrap.c so that it will clean up a
   stale UNIX socket. I opened an upstream PR too but I'm not sure if the code is
   maintained.

10. Installing the package gave this warning:

        FMRI pattern might implicitly match more than one service instance.
        Actuators for refresh_fmri will not be run for
        svc:ooce/application/fcgiwrap.

   so the FMRI needed qualifying (:// at start, :default at end)

11. I allocated 88 for the uid and gid. I've opened a new PR to add `tools/idlist` to show all of
     the IDs that are allocated by extra packages. Of course there are a few used by core packages
     and by illumos itself to be taken into account, but it's good to have explicit numbers in the package
     for consistency, and 88 is the next free number.

pbdigital-dot-org:
great! I will sort this out tomorrow morning.

hadfl:
since `ooce/library/fcgi2` is a build-time requirement of this package. we should add that to the `extra-build-tools` meta package.

```diff
--- a/build/meta/extra-build-tools.p5m
+++ b/build/meta/extra-build-tools.p5m
@@ -17,6 +17,7 @@ depend fmri=ooce/developer/llvm-80 type=require
 depend fmri=ooce/developer/llvm-90 type=require
 depend fmri=ooce/developer/rust type=require
 depend fmri=ooce/library/cairo type=require
+depend fmri=ooce/library/fcgi2 type=require
 depend fmri=ooce/library/libev type=require
 depend fmri=ooce/library/libexif type=require
 depend fmri=ooce/library/libid3tag type=require
```

pbdigital-dot-org:
The fuzz on the patch file was "my bad". I initially patched against a github tag release and then did not update the patch when I changed the source to the 1.1.0 release.

citrus-it:
> The fuzz on the patch file was "my bad". I initially patched against a github tag release and then did not update the patch when I changed the source to the 1.1.0 release.

No problem, a lot about this framework is not obvious and we are lacking documentation. This gave an opportunity to point out `./build.sh -P` which we usually use when updating to a newer version of any package.

Thanks again for these  PRs and for taking the time you obviously have to fit in with the framework.

citrus-it:
Could you add the `fgci2` package to `build/meta/extra-build-tools.p5m` please?
That's a meta package which should install everything that's needed to rebuild the whole of the extra repository. There are obvious bootstrapping issues but we'll shortly get fcgi2 published for supported releases.

pbdigital-dot-org:
@hadfl , if you like I can update the `extra-build-tools` meta package. However, it will be this evening.
@citrus-it , you're welcome. thanks for your patience.
Philip

citrus-it:
> @hadfl , if you like I can update the `extra-build-tools` meta package. However, it will be this evening.

No problem, I'll do that and get this merged.

pbdigital-dot-org:
@citrus-it , I just did it

hadfl:
Thank you, looking forward to your next contribution!

PR #480, opened 2020-04-17
New package zonetools

cneira:
I would like to include this package in extras, the package is a couple of scripts to check
zone memory usage and disk activitiy.

Man pages and scripts are located in  https://github.com/cneira/zonetools

### DIFF
--- a/build/zonetools/build.sh
+++ b/build/zonetools/build.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/bash
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License, Version 1.0 only
+# (the "License").  You may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


citrus-it:
Thanks @cneira these look really useful. Am I right in thinking the originals are in the illumos-joyent tree? If so, I would prefer to take them into illumos-omnios from there and then we would track any future updates that way.

cneira:
> Thanks @cneira these look really useful. Am I right in thinking the originals are in the illumos-joyent tree? If so, I would prefer to take them into illumos-omnios from there and then we would track any future updates that way.

Yes, here is zonememstat https://github.com/joyent/smartos-live/blob/c9298c50c8ee4c7e3b80e434cc60b5afdf10a310/src/zonememstat.sh

I recall there is small modification to do to run it on Omniosce.

PR #483, opened 2020-04-23
add nagios-4.4.5 build files

pbdigital-dot-org:
tarball info:
tarball: https://github.com/NagiosEnterprises/nagioscore/releases/download/nagios-4.4.5/nagios-4.4.5.tar.gz
( sha256 bf69e2d2fae218694585677472d355ba676120cbd024164281e635dc467c391d )
extracts to nagios-4.4.5

build info:
pkg install git gcc8 gnu-make ripgrep pigz gnu-tar autoconf automake libtool pkg-config

notes in local.mog
1. only made bin symlinks, so as not to fill up sbin with lots of cgi files
2. changed lots of permissions as this is how nagios normally installs, I kept /opt/ooce/nagios as root bin but had to change these nagios binaries to 755 as they are normally set 744
3. also created group nagcmd as is normal in nagios installation
4. used uid/gid of 120/1 ,as it is pretty crowded in the 90's with system uid's

notes on build.sh
1. MAKE_INSTALL_ARGS: nagios will try to install files as nagios_user/group but package is built before these users are created I set the variable to empty so files can be installed

I also added some helper files under the files directory.

### DIFF
--- a/build/nagios/build.sh
+++ b/build/nagios/build.sh
@@ -47,7 +47,7 @@ MAKE_INSTALL_ARGS="
 
 MAKE_INSTALL_TARGET="
     install
-    install-commandmode 
+    install-commandmode
     install-config
 "
 
@@ -57,7 +57,7 @@ XFORM_ARGS="
     -DPROG=$PROG
 "
 
-CONFIGURE_OPTS_64=" 
+CONFIGURE_OPTS_64="
     --prefix=$PREFIX
     --sysconfdir=/etc$PREFIX
     --localstatedir=/var$PREFIX
@@ -70,6 +70,7 @@ patch_source
 prep_build
 build
 strip_install
+add_notes README.install
 install_smf application $PROG.xml application-$PROG
 make_package
 clean_up
--- a/build/nagios/build.sh
+++ b/build/nagios/build.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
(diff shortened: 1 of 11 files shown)


citrus-it:
@pbdigital-dot-org - thanks a lot for these two PRs. There's quite a bit to look at here, you raise some interesting questions that are going to need a bit more research and we're in the final stages of testing the upcoming OmniOS r151034 release so please bear with us as it will take a few days.

pbdigital-dot-org:
> in the final stages of testing the upcoming OmniOS r151034 release so please bear with us as it will take a few days.

no problem - looking forward to the r151034 release

citrus-it:
I don't have many comments on this one at all, it's really good; thanks!

### UIDs

This always presents a challenge and I wish that OmniOS had started with users numbered from 1000. We had a chat about this and decided that the best option is to continue to allocate specific numbers from the reserved (<100) range. If a number is not specified in the manifest, IPS is capable of just picking a number (one above the current maximum) but that can lead to differences between systems. The problem with choosing a number like 120 is that it could be a real user on systems that want to install this package.

We've allocated `83` & `84` for this package to use.

### installation notes

For something like nagios which requires manual integration steps, providing a README as you have done is the right thing to do. IPS has a facility to display notes during package installation so I've added a short one (edit as you require) which points to the full file which is now installed under `/opt/ooce/nginx/share/`.

### Example configuration file

I added a line to local.mog to install this file example under /etc/opt/ooce/nginx/

### Method script

The check for whether Nagios is already running was not really robust. It only checked to see if there was any process with that process ID. I've changed it to check that the process is actually nagios.

One way that stale PID files often occur is through the normal upgrade process where a new boot environment is created. The old boot environment is cloned while processes are running which means that PID files etc. are copied over. When the new BE is booted, it can cause services to fail to start.

(In general, we'd recommend not using a method script at all and allowing SMF to manage all of the processes under a service contract. It certainly cleans up the logic around making sure that all processes are cleaned up when the service is disabled etc. - this is not always possible of course)


You can see my suggested changes in this commit - I did fix some trailing whitespace too but there are not many changes.

https://github.com/omniosorg/omnios-extra/commit/3445d7dbe53b29f2ec428f3c61908b511c6858b3

Thanks again for the contribution, I'm going to review the nagios-plugins PR next.

PR #484, opened 2020-04-23
add nagios-plugins-2.3.3 build files

pbdigital-dot-org:
tarball: http://www.nagios-plugins.org/download/nagios-plugins-2.3.3.tar.gz
( sha256 07859071632ded58c5135d613438137022232da75f8bdc1687f3f75da2fe597f )
extracts to nagios-plugins-2.3.3

notes on build.sh
1. RUN_DEPENDS_IPS: was not sure how to make build dependent packages optional for run_depends. looking under functions.sh I saw it possibe to "case" with variable symbols as the first character. 
however, this did not remove the original "required" dependency so I added the following line in my functions.sh to drop the required package in the final.mog

```
             #functions.sh - lines 1318-1320
             if grep -q "^depend .*fmri=[^ ]*$depname" "${P5M_INT3}.res"; then
+                logcmd sed -i "\%^depend .*fmri=[^ ]*$depname%d" ${P5M_INT3}.res
                 autoresolved=true
```

without this the build will have duplicate dependencies. one required, the other optional.

just a note, check_ldap was not built as it depends on lldap and lber. ldap is there. libber is not, however `/usr/include/liber.h` exists.

### DIFF
--- a/build/nagios-plugins/build.sh
+++ b/build/nagios-plugins/build.sh
@@ -42,7 +42,7 @@ MAKE_INSTALL_TARGET="
     install-root
 "
 
-CONFIGURE_OPTS_64=" 
+CONFIGURE_OPTS_64="
     --prefix=$PREFIX
 "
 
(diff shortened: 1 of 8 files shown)


citrus-it:
This commit shows how to remove the detected `require` dependencies:

https://github.com/omniosorg/omnios-extra/commit/0c88eb1f1a058e0289eb695e0749c6ed9808530e

I also changed the download location since we'll put the plugins alongside nagios on the mirror.


FWIW, optional dependencies specified without a version component are effectively no-ops, but there is no harm them being present in the manifest.

pbdigital-dot-org:
Thanks Andy

PR #498, opened 2020-05-08
Nagios which relies on nagios common

pbdigital-dot-org:


build/nagios/build.sh
```diff
@@ -33,6 +33,14 @@ all be accessed via a web browser."
 
 set_arch 64
 
+BUILD_DEPENDS_IPS="
+    ooce/application/nagios-common
+"
+
```
citrus-it, line 39: Does this package definitely need nagios-common installed on the build system in order to build?
If not, please remove this block.
If yes, then add nagios-common to extra-build-packages

pbdigital-dot-org: I have just rebuilt these with out depending on the common package so will drop the blocks from `build.sh` and make a commit soon.

PR #499, opened 2020-05-08
Nagios plugins which relies on nagios common

pbdigital-dot-org:


build/nagios-plugins/build.sh
```diff
@@ -27,11 +27,13 @@ set_arch 64
 BUILD_DEPENDS_IPS+="
     ooce/database/mariadb-104
     ooce/database/postgresql-12
+    ooce/application/nagios-common
```
citrus-it, line 30: As per the other package, I don't think nagios-common needs to be installed in order to build this package.

pbdigital-dot-org: I have just rebuilt these with out depending on the common package so will drop the blocks from `build.sh` and make a commit soon.

PR #507, opened 2020-05-13
sysstat: Add -n option in configure ksh script to prevent build error

v-a-b:
The configure script executes two branches. In one, the getopts parameters string doesn't contain -n, thus triggering an error when "configure -n" is called.


citrus-it:
Some further background - the `mkrules` target in the Makefile calls `./configure -n` and in a ksh93 environment, this is not passed to getopt as a valid option.

PR #516, opened 2020-05-25
add tcpdump build files

pbdigital-dot-org:
Hi,
I am aware `snoop` is in the base install but I thought `tcpdump` may be useful to those that are accustomed to it. 
All the tests have passed and are in the `testsuite.log` file.

Philip


build/tcpdump/build.sh
```diff
@@ -0,0 +1,41 @@
...
+# http://www.illumos.org/license/CDDL.
+# }}}
+
+# Copyright 2020 OmniOS Community Edition.  All rights reserved.
+
+. ../../lib/functions.sh
+
+PROG=tcpdump
+VER=4.9.3
+PKG=ooce/network/test/tcpdump
```
citrus-it, line 21:
```suggestion
PKG=ooce/network/tcpdump
```
Let's drop the `/test` component

Otherwise, looks perfect, thanks!

pbdigital-dot-org: cheers! 
also, I noticed `socat` was missing from `doc/baseline` so I have added
Philip

citrus-it: `socat`  is a strange one in that it used to be in the extra repository but it was moved to core in r151031 - see https://github.com/omniosorg/omnios-extra/blob/master/build/socat/build.sh#L26

baseline tracks master which is why it's missing there - otherwise when doing a full extra build on anything newer than r151031, you get an error about a mismatched baseline.

pbdigital-dot-org: thanks, I will keep an eye out for things like that in the future

PR #534, opened 2020-06-10
packaging README.md

pbdigital-dot-org:
This is almost done, I have a little work to do on the "specific packages" section but thought it would be a good time to get it up now.
In the package example section, I created a helloworld source tarball that builds under OmniOS and think it will be less confusing to use that as the example. I have used `set_mirror` in the `build.sh` to download the source but I think it would be better if we could remove this, so that we have a clean `build.sh` template in the document. Though, that would mean the helloworld source tarball would need to go on the omniosce mirror, if that is OK.
Let me know what you think,
Philip

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The "OmniOS Extra IPS Repository" is open to all, anyone may submit a new packag
 
 1. [Requirements](#requirements)
 2. [Create the build environment](#create-the-build-environment)
-3. [Create a build directory and it's build files](#create-a-build-directory-and-it-s-build-files)
+3. [Create a build directory and it's build files](#create-a-build-directory-and-its-build-files)
 4. [Build the `helloworld` package](#build-the-helloworld-package)
 
 ### The "OmniOS Extra Build System"
@@ -30,11 +30,12 @@ The "OmniOS Extra IPS Repository" is open to all, anyone may submit a new packag
 11. [Providing messages at installation](#providing-messages-at-installation)
 12. [Making the source available to the build environment](#making-the-source-available-to-the-build-environment)
 13. [Managing licences](#managing-licences)
-14. [The build system's default "IPS Repository"](#the-build-system-s-default-ips-repository)
+14. [Adding package information to the build system](#adding-package-information-to-the-build-system)
+15. [The build system's default "IPS Repository"](#the-build-system-s-default-ips-repository)
 
 ### Tips for building specific types of packages
 
-1. [Building packages that share Common elements](#building-packages-that-share-common-elements)
+1. [Building packages that share common elements](#building-packages-that-share-common-elements)
 2. [Tips for building libraries](#tips-for-building-libraries)
 3. [Tips for Go packages](#tips-for-go-packages)
 4. [Tips for Perl packages](#tips-for-perl-packages)
@@ -55,7 +56,7 @@ The objective of this section is to prepare a build environment and introduce th
 
 1. [Requirements](#requirements)
 2. [Create the build environment](#create-the-build-environment)
-3. [Create a build directory and it's build files](#create-a-build-directory-and-it-s-build-files)
+3. [Create a build directory and it's build files](#create-a-build-directory-and-its-build-files)
 4. [Build the `helloworld` package](#build-the-helloworld-package)
 
 ## Requirements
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1 +1,922 @@
...
+Sometimes the source tarball is not in a standard format and the argument to the directive, `download_source`, will need to be edited accordingly.
+
+#### Build Directory
+
+Sometimes the tarball will extract to an un-expected directory. For example, sometimes a tarball named `$PROG-$VER.tar.gz` may extract to a directory named `$PROG-$PROG-$VER`.
+
+In this case, use the `BUILDDIR` directive, as follows:
+
+```none
+BUILDIR="$PROG-$PROG-$VER"
```
citrus-it, line 683: Use `set_builddir "$PROG-$PROG-$VER"`

README.md
```diff
@@ -1 +1,922 @@
...
+```
+
+#### Source Directory
+
+Sometimes the source directory, where the `configure` script lies will not be where it is expected. For example, the `configure` script may not be in the root of the `$PROG-VER/` directory, but instead in `$PROG-VER/$PROG/`.
+
+In this case, use the `“EXTRACTED_SRC` directive, as follows:
+
+```none
+EXTRACTED_SRC+="/$PROG"`
```
citrus-it, line 693: You should never need to do this. `set_builddir`, see above, takes care of EXTRACTED_SRC too

README.md
```diff
@@ -1 +1,922 @@
...
+
+```none
+EXTRACTED_SRC+="/$PROG"` 
+```
+
+## Managing licences
+
+### Bundling a licence with the package
+
+All packages must be bundled with thier respective licence. To add a licence to the package being built, the licence must be included in the `local.mog` file. This is generally always the last line of text in the `local.mog` file and appears as follows:
```
citrus-it, line 700: s/thier/their/

README.md
```diff
@@ -1 +1,922 @@
...
+    sed '/^$/q' < $PROG.c > LICENCE
+    popd > /dev/null
+}
+```
+
+The licence may then be determined by the `tools/licence` utility.
+
+### When all else fails
+
+As a last resort, you may consider mofifying the licence definition pattern in `doc/licences`. **You will need to let the maintaners explicitly aware of any changes to this file.**
```
citrus-it, line 761: Typo `mofifying` and `maintaners`

citrus-it:
> Would you like feedback on the documentation itself at all? I've had building and uploading my first package on my todo list for a while and will happily suggest any changes or additions from the point of view of someone new to the IPS system if it would help the PR :)

Definitely - you can't beat updates and observations from a fresh pair of eyes.
This will hopefully get merged in the next couple of days so you can raise a new pull request against the document then, or put comments in an issue.

Thanks!

pbdigital-dot-org:
@citrus-it I have resolved all those comments. And completed the 2 sections that were blank. Regards, the other sections "Tips for xxxx packages", these can be place holders for now, and others with more expertise in this area can look at updating them.
And in time, for sure, there will be more sections to add.

README.md
```diff
@@ -1 +1,961 @@
...
+
+To create packages for the "OmniOS Extra IPS Repository", the following is necessary:
+
+* Virtual or  physical system running the latest OmniOS release, preferably the  latest ["Bloody Release"](https://omniosce.org/about/stablevsbloody).
+* The system should have a minimum of 2GB of RAM.
+
+* "OmniOS [Build Tools](https://github.com/omniosorg/omnios-build/blob/master/build/meta/omnios-build-tools.p5m)": this can be installed with the following command:
+
+```none
+# pkg install ooce/omnios-build
```
citrus-it, line 72: It's just `ooce/omnios-build-tools` that's required, you removed the wrong one, and lost the `-tools` on the install command.

pbdigital-dot-org: I'm confused: That link points to omnios-build-tools. That is the only required package, no? 
The command I get, I can add the trailing `-tools`  once the above is confirmed.

citrus-it: Sorry, I made a mistake too.

The only package you need to install for working on extra, is `ooce/extra-build-tools`
https://github.com/omniosorg/omnios-extra/blob/master/build/meta/extra-build-tools.p5m

This will automatically pull in everything that's required.

pbdigital-dot-org: that is sorted.

PR #538, opened 2020-06-16
docs/{baseline,packages.md} should be doc/...

ptrrkssn:
Very basic fix for the README.md file - docs/ should be doc/ (for baseline & packages.md)
- PE

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ The main engine behind the "OmniOS Extra Build System" is the `lib/functions.sh`
 
 ### Package catagories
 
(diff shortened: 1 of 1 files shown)


ptrrkssn:
A second comment regarding "packages.md": 

I'm not 100% sure what the last entries per line in this file should be. The:

  ... [omniosorg](https://github.com/omniosorg) 

part when adding new packages. Copy that verbatim or change it to something else ([my-user-name](url-to-website)) ? 

Perhaps some clarification on that could be added to README.md?

PR #539, opened 2020-06-16
Add a section on a possible use of "local" tarballs

ptrrkssn:
A suggestion for a new section in the README.md file on how to use "local" tarballs. It's a bit of a hack but seems to work for me at least. The wording can probably be _much_ improved. But it's nice to be able to test the build process locally without having to publish a release on Github every time (especially when working on the original source and trying to fix it so it is portable to multiple OS:es at the same time... I typically test my code on FreeBSD, OmniOS, Linux and MacOS.

(I got the idea from the MacOS HomeBrew build process where this was possible)

- Peter

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,6 +677,19 @@ GitHub does not supply checksums for downloaded archives, so this can be workd a
 set_checksum "none"
 ```
 
+### Local source tarballs
+
+Sometimes it might be useful to be able to use "local" tarballs without having to go via a Mirror or a GitHub release.
+For example when doing development of the original source and wanting to test the OmniOSCE Extra build process without
+having to publish a temporary "release" every time.
+
+This can be accomplished by preparing a distribution tarball of the project and after having configured stuff per the
+instructions above (for a Mirror or GitHub release case) then - just before you run "./build.sh" - copy the tarball to
+the directory /tmp/build_$USER/$PACKAGE-$VERSION/ (see TMPDIR in lib/config.sh for the exact location of /tmp/build_$USER/).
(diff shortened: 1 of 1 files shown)


oetiker:
hmmm wouldn't it make more sense (if this is not the case yet) to just update build.sh such that it can handle file:// resources ?

ptrrkssn:
Mm.. Perhaps. However with “my way” the idea is that “build.sh” can be set up exactly the way it is supposed to be in the end (getting the tarball from the official source) - so it will work automatically when the real distribution tarball finally is made available there - I just skip “preloading” the tarball. 

This probably only is useful for people like me though when doing development both on the OmniOS “port/package” _and_ then original source code at the same time :-)

(I find it useful to test-drive program builds and package generation on different platforms more or less concurrently - it weeds out different bugs and differences in an interesting (albeit frustrating sometimes :-) way. Sometime small (like the fact that FreeBSD “ports” really want installed binaries to be stripped - and OmniOS likes them _not_ to be stripped) - and sometimes bigger). 

A file:/// <file:///> source could also be useful though, but for other reasons.

- Peter

citrus-it:
Actually, you can just `set_mirror /path/to/tarfiles` or set `MIRROR` in your local `lib/site.sh` file.
I think using the latter technique is closer to what you want as the build script remains unchanged.

Either way, the mirror can already point at a local directory.

(Re: stripping - we don't strip symbol tables which means using `strip -x`, and ideally we like CTF data in the final binaries as it eases debugging without significantly increasing the size

`build -ctf`

takes care of adding CTF data and removing almost everything else in most cases, and the default

`strip_install`

action strips without removing the symbol table.

PR #542, opened 2020-06-20
Enable package-local rtime file for overrides

ptrrkssn:
This patch adds support for package-local "rtime" files to override/add exceptions defined in "doc/rtime" file on a per-package basis. This can be useful if a package has special needs.

One such example is libreadline which must be linked with either libtermcap, libcurses or libncurses - but neither one of then is referenced by libreadline - so the check_rtime tool will always complain about libtermcap/libcurses/libncurses been unneeded. With this patch one can add a local "rtime" file in the package directory like this:

```
# libreadline needs either one of libtermcap | libcurses | libncurses,
# but is not linked with either one of them
UNREF_OBJ libtermcap\.so\.1
UNREF_OBJ libcurses\.so\.1
UNREF_OBJ libncurses\.so\.6
```

### DIFF
--- a/build/acltool/rtime
+++ b/build/acltool/rtime
@@ -1,0 +1,7 @@
+
+# libreadline needs either one of libtermcap | libcurses | libncurses,
+# but is not linked with either one of them
+UNREF_OBJ libtermcap\.so\.1
+UNREF_OBJ libcurses\.so\.1
+UNREF_OBJ libncurses\.so\.6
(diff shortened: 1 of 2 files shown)


citrus-it:
I opened https://github.com/omniosorg/omnios-build/pull/1914 to fix the underlying problem with the readline library. That should be integrated into bloody soon.

PR #557, opened 2020-07-08
add apr-util package

pbdigital-dot-org:


build/meta/extra-build-tools.p5m
```diff
@@ -20,6 +20,7 @@ depend fmri=ooce/developer/gperf type=require
 depend fmri=ooce/developer/llvm-90 type=require
 depend fmri=ooce/developer/llvm-100 type=require
 depend fmri=ooce/developer/rust type=require
+depend fmri=ooce/library/apr-util type=require
```
citrus-it, line 23: This should be just `ooce/library/apr`

PR #558, opened 2020-07-08
add subversion package

pbdigital-dot-org:


pbdigital-dot-org:
yes, that should all be good to go

PR #572, opened 2020-07-24
X11

hadfl:
getting prerequisites in for building `openjdk11`


danmcd:
Dumb question, is this enough X to actually run, say twm on one's HDC console?  Or is it less than that (literally just enough to satisfy openjdk 11)?

oetiker:
@danmcd since there is no x server I guess it is not enough for the console ... :)

PR #587, opened 2020-08-09
BIND 9.11.21 Server

pbdigital-dot-org:


pbdigital-dot-org:
I made a couple of additions:
In the `build.sh`, I have allowed for the _Locked Hardlinks_. 
I added to `named.conf` the option to use flat file logging, and added the respective directory via `local.mog`. `syslog` is still the default.
Regards "forwarding configured automatically based on the entries in `/etc/resolv.conf`", this is specific to the FreeBSD rc/init/daemon script so can be safely disregarded.
I have disabled remote administration via the `control` clause in `named.conf`. This clears up the control error messages, and can be configured if needed.

PR #617, opened 2020-09-03
add manifest tidy tool

pbdigital-dot-org:
OK, this is a starting point...
does not format quotes as this may need manual inspection for nested quotes.
formatting for license indentation is needed, will look at adding this tomorrow.
At the moment I am not familiar with which base directories to use etc, tomorrow I will inspect the other `tools/` scripts and update this script. 
For the moment, this will give us formatted manifests and then the conf file can be discussed with which preferences are desired.

### DIFF
--- a/tools/tidy-manifests
+++ b/tools/tidy-manifests
@@ -1,0 +1,55 @@
+#!/bin/ksh
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+
+# Copyright 2020 OmniOS Community Edition (OmniOSce) Association.
+
+tidy=/opt/ooce/bin/tidy
+__SCRIPTDIR="${0%/*}"
+
+ELEMENTS=(
+        '<service '
+        '<create_default_instance'
+        '<single_instance'
+        '<instance'
+        '<dependency'
+        '<dependent'
+        '<exec_method'
+        '<property_group'
+        '<template'
+        '<stability'
+        '<\/service'
+        '<\/instance'
+)
+
+echo "PROCESSING: $1"
+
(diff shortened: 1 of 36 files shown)


citrus-it:
I'm happy with this as it stands - the output format isn't one of the ones I'd have thought of but it looks ok and it means we can get consistent manifest files.

Given that this is running against a git checkout, the backup file is probably not necessary.

pbdigital-dot-org:
> Given that this is running against a git checkout, the backup file is probably not necessary.
I was thinking that if this is run whilst creating the package, before any commits. If you think it does not belong, I can drop it no problem.
Later I will go through each manifest separately to make sure I have all the elements that need additional spacing and look for nested quotes.

hadfl (approved):

oetiker (dismissed):

tools/tidy-manifests
```diff
@@ -0,0 +1,42 @@
...
+        '<template'
+        '<stability'
+        '<\/service'
+        '<\/instance'
+)
+
+echo "PROCESSING: $1"
+
+# tidy manifest according to doc/tidy.conf
+$tidy -q -config ./doc/tidy.conf -m $1
```
citrus-it, line 23: Could you add something to this script so that it doesn't need to be run from the top-level of the checkout?
Often if I'm working on a package I'll be in `build/xxx` and want to be able to do `../../tools/tidy-manifest files/xxx.xml`

Something like
```
__SCRIPTDIR="${0%/*}"
```
then
```
$__SCRIPTDIR/../doc/tidy.conf
```
will work for ksh (if you change the shebang)

citrus-it (commented):
Can you use this to update all of the manifests, and add that to this PR as a separate commit?
That will let us review the output that the script produces too.

tools/tidy-manifests
```diff
@@ -0,0 +1,42 @@
...
+if [ $? -gt 0 ]; then
+    echo "An error occurred processing $1!"
+    exit 0
+fi
+
+# use only double quoutes in manifest and chceck for nested quotes
```
citrus-it, line 29: Just a couple of typos here
```suggestion
# use only double quotes in manifest and check for nested quotes
```

citrus-it (approved):

PR #629, opened 2020-09-14
/etc/inet/services fragments

citrus-it:


oetiker (approved):

oetiker:
the user has to take care to not produce conflicts, right ?

citrus-it:
> the user has to take care to not produce conflicts, right ?

No, the assembly script should resolve/merge conflicts and combine aliases etc.

(For example, NNTP is here and also in the standard services file - the list in this PR is straight from the cyrus packaging guide)

hadfl (approved):

PR #635, opened 2020-09-16
Split postgres12 into client and server package.

mvwieringen:
First stab at splitting PostgreSQL 12. Same concept as mariadb with the same problems.
Just as heads up needs more work.


mvwieringen:
New version based on the new scripting. First test look promising already has support for packaging both 32/64 bits when
that gets added later on.

PR #640, opened 2020-09-17
Build postfix with sqlite3, mariadb and postgresql support

mvwieringen:
Given we now depend on cyrus sasl libs which depends on mariadb and PostgreSQL this change is low hanging fruit that
has little impact on the postfix package. The only changes are:

+depend type=require fmri=pkg:/ooce/library/mariadb-104
+depend type=require fmri=pkg:/ooce/library/postgresql-12

for the newly published package. I myself don't use mysql or postgresql tables in my postfix installs but people might
and we can give them what they want for no additional costs other than these small build changes.

### DIFF
--- a/build/postfix/build.sh
+++ b/build/postfix/build.sh
@@ -40,6 +40,8 @@ BUILD_DEPENDS_IPS="
     library/pcre
     ooce/database/bdb
     ooce/database/lmdb
+    ooce/library/postgresql-${PGSQLVER//./}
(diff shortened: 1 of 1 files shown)


hadfl (dismissed):

citrus-it (approved):

oetiker (approved):

mvwieringen:
b.t.w. the only change between the first version is that I squashed the two patches to one and changed the commit message.

PR #649, opened 2020-09-24
Add postgresql 13

citrus-it:


mvwieringen (changes requested):
From a quick check of the client package your new patterns are somewhat to loose e.g. it also selects opt/ooce/pgsql-13/lib/libpqwalreceiver.so which should not be in the client package. This is something I missed in reviewing last time. My old pattern had this to be sure that one didn't get selected:

manifest_add $PREFIX/lib 'libpq\..*' 'libecpg.*' 'libpgtypes.*' e.g. libpq\. to be sure we skip libpqwalreceiver

citrus-it:
> From a quick check of the client package your new patterns are somewhat to loose e.g. it also selects opt/ooce/pgsql-13/lib/libpqwalreceiver.so which should not be in the client package. This is something I missed in reviewing last time. My old pattern had this to be sure that one didn't get selected:
> 
> manifest_add $PREFIX/lib 'libpq.._' 'libecpg._' 'libpgtypes.*' e.g. libpq. to be sure we skip libpqwalreceiver

Thaks, fixed

mvwieringen:
> > From a quick check of the client package your new patterns are somewhat to loose e.g. it also selects opt/ooce/pgsql-13/lib/libpqwalreceiver.so which should not be in the client package. This is something I missed in reviewing last time. My old pattern had this to be sure that one didn't get selected:
> > manifest_add $PREFIX/lib 'libpq.._' 'libecpg._' 'libpgtypes.*' e.g. libpq. to be sure we skip libpqwalreceiver
> 
> Thaks, fixed

Ok think we have the same problem in pgsql 12 as I didn't review the previous version good enough e.g. seems you optimized it a bit to much and I was to blind to catch that.

mvwieringen:
Ignore that last comment seems both 12 and 13 are ok.

citrus-it:
Yes, I updated both 12 and 13 in the last push.

mvwieringen (dismissed):

oetiker (approved):

oetiker:
is there a particular reason for dropping support for 10 ? EOL ?

citrus-it:
> is there a particular reason for dropping support for 10 ? EOL ?

I was wrongly just keeping the past three main packages in the repository, but in case users are still running version 10 we should carry on publishing it. It is not EOL until 2022

Instead of dropping it, I've marked it `legacy`

mvwieringen (approved):

hadfl (approved):

PR #652, opened 2020-09-25
Add mutt package

citrus-it:


hadfl (approved):

danmcd (approved):
Oooh, I'll definitely be using this!

PR #718, opened 2020-11-20
Library updates

pbdigital-dot-org:
Regards `protobuf` library, there is an ABI change, as follows:

`libprotoc ABI change, 24  -> 25 `

It looks like just `mosh` that uses `protobuf`, therefore I have built `mosh` with new `protobuf` and the build completes without errors.

### DIFF
--- a/build/libjpeg-turbo/build.sh
+++ b/build/libjpeg-turbo/build.sh
@@ -17,7 +17,7 @@
 . ../../lib/functions.sh
 
 PROG=libjpeg-turbo
-VER=2.0.5
+VER=2.0.6
 PKG=ooce/library/libjpeg-turbo
 SUMMARY="libjpeg-turbo"
 DESC="SIMD-accelerated libjpeg-compatible JPEG codec library"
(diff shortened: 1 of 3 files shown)


citrus-it (approved):

citrus-it:
> Regards `protobuf` library, there is an ABI change, as follows:
> 
> `libprotoc ABI change, 24 -> 25 `
> 
> It looks like just `mosh` that uses `protobuf`, therefore I have built `mosh` with new `protobuf` and the build completes without errors.

We generally have to be careful because people could have built their own software with a dependency on this and then `libprotoc.so.24` disappears under them. However, we have not been keeping the old ABIs around for protobuf and there have been no complaints, so we will just have to make sure we rebuild mosh too .

pbdigital-dot-org:
> We generally have to be careful because people could have built their own software with a dependency on this and then `libprotoc.so.24` disappears under them. However, we have not been keeping the old ABIs around for protobuf and there have been no complaints, so we will just have to make sure we rebuild mosh too .

thanks for feedback. I will make sure to flag any of these with ABI changes. 
Is there any library in extra that tracks multiple ABIs so I can have a look at the build files to see how this is done

hadfl (approved):

PR #754, opened 2020-12-16
Move PHP FPM socket back to previous location

citrus-it:


hadfl (approved):

beiDei8z:
Can we please stop moving files without test?  These are for the Long Term Stable branch too.

To be consistent, if prefix=/opt/ooce/php-$version, then var is /var/$prefix and certainly not two locations.
 
If files are moved then fill the gaps with symbolic links.

citrus-it:
The socket file was inadvertently moved in #752 along with the FPM log file and it should not have been. This moves it back to where it has always been historically to avoid breaking existing setups. Most people won't be affected unless they installed the broken package update in the small window yesterday and, for them, another update will fix it.

In contrast with the core OmniOS OS,  these extra packages don't have an LTS branch (or any branches, there is just master). Rolling package updates are published for all OmniOS releases which still have support and one can control updates with `pkg freeze` if desired.

PR #764, opened 2020-12-28
gnu-st: GNU smalltalk 3.2.5

cstes:
This package was originally written by Steve Byrne who was a Sun employee I think, longtime ago ...

I can build it in the omnios-extra and it installs ok for me in the https://pkg.omniosce.org/r151036/extra/

Please note that there are some issues where I just type "yes" when it asks "do you want to continue" ?

For example,

1) unused object=/lib/libpthread.so.1      <remove lib or -zignore?>
ELF runtime problems detected

2) -- Checking for library ABI changes
Could not retrieve contents

3) These hardlinks do not have locked targets, resulting in inconsistent builds.
--- Unlocked hardlink: opt/ooce/bin/gst-load <- opt/ooce/bin/gst-blox

However when I just go ahead

===== User elected to continue after prompt. =====

Then it stilll seems to build fine and install OK for me.

When I manually build GNU smalltalk the testsuite reports 2 failures

gmake check testsuite

ERROR: 129 tests were run,
2 failed (1 expected failure).
2 tests were skipped.

## ------------------------ ##
Failed tests:
GNU Smalltalk 3.2.5 test suite test groups:

In the omnios-extra build, there are 3 failures in the testuite , I believe.

This may be due to more agressive compiler optimizer flags in the omnios-extra build,
but in any case most of the testsuite (run_testsuite) is succesful.

Please consider this pull request for merge.

Regards,
David Stes

### DIFF
--- a/build/gnu-smalltalk/build.sh
+++ b/build/gnu-smalltalk/build.sh
@@ -1,0 +1,90 @@
+#!/usr/bin/bash
+#
+# {{{
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+#
+# Copyright 2020 David Stes
+# GNU Smalltalk latest stable 3.2.5 with some patches from 3.2.91 alpha
+
+. ../../lib/functions.sh
+
+PROG=gnu-smalltalk
+VER=3.2.5
+PKG=ooce/runtime/gnu-smalltalk
+SUMMARY="GNU Smalltalk"
+DESC="A free implementation of the Smalltalk-80 language, \
+well-versed to scripting tasks and headless processing."
+
+# 64-bit build currently segfaults, needs investigation
+set_arch 32
+set_builddir smalltalk-$VER
+
+HARDLINK_TARGETS="
+    opt/ooce/gnu-smalltalk/bin/gst-load
+"
+
+OPREFIX=$PREFIX
+PREFIX+=/$PROG
+
+reset_configure_opts
+
+TESTSUITE_FILTER='^[A-Z#0-9 ][A-Z#0-9 ]'
+
+# despite the fact that GNU smalltalk uses its own libtool
+# there is a need for a ltdl.h header file, so require libtool
+# GNU Smalltalk can be compiled --without-emacs but conceptually,
+# it seems better to require it as the GNU smalltalk integrates with Emacs
+BUILD_DEPENDS_IPS+="
+    developer/build/libtool
+    ooce/editor/emacs
+"
+
+# deliberately force emacs as dependency;
+# GNU Smalltalk supports emacs and emacs as its IDE
+
+RUN_DEPENDS_IPS+="
+    ooce/editor/emacs
+"
+
+XFORM_ARGS="
+    -DOPREFIX=${OPREFIX#/}
+    -DPREFIX=${PREFIX#/}
+    -DPROG=$PROG
+    -DPKGROOT=$PROG
+"
+
+# disable-generational-gc because libsigsegv not found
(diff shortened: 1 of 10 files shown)


hadfl:
Hi David, thank you very much for your contribution. Sorry that it's been taking so long to review this but it's holiday season so there's a lot of family business going on.

First of all: The PR looks pretty good.
There are just a couple of changes where I'd like your feedback. I've drafted a commit that sums them up:
https://github.com/omniosorg/omnios-extra/commit/18dac460fbc11c497efbc1a7037a29f77e7fc423

It boils down to:
- the shebang should be on the first line
- renaming the package from `gnu-st` to `gnu-smalltalk` so other users are more likely to find it with `pkg search`
- I've put the source on our mirror so we can drop specifying the mirror or checksum
- defining the hardlink target so we get a consistent pkgdiff
- moving the package contents from `/opt/ooce` to `/opt/ooce/smalltalk` as we tend to only deliver dual arch libraries directly to `/opt/ooce`
- adding a filter to the test suite results so the diff with new versions is easier to read
- adding the package to `doc/packages.md` and `doc/baseline` (also adding emacs to `extra-build-tools` meta package as it is now a build time requirement to build omnios-extra)

One thing I am not sure about is whether it is worth to disable optimisation at all. It seems that the exceptions test fails when optimisations are enabled. I think it's rather worth to keep optimisation enabled but maybe investigate why the exceptions test fails. I'd really like your opinion on this matter.

Please have a look at the proposed changes. If you agree I'll squash them with your commits (while keeping you as the author) and get this ready for integration.

dcstes:
Thanks for your feedback, and for having a look at this.    I would like to do some GNU smalltalk software development on OmniOS and therefore propose this (old) package which is (or was?) included for many years in Ubuntu/Debian Linux : https://packages.debian.org/bullseye/gnu-smalltalk.   The name they choose in Debian, is gnu-smalltalk although the official name is simpy 'smalltalk'.  However that is a bit unfortunate since there exist different implementations.  So gnu-smalltalk is a good choice,  that name is OK.   Regarding the -O2 issue I can try to have a look and debug the issue although I cannot promise I'll find anything or give a timeframe for that.  But I'll try and come back later on this.  Regards, David Stes

hadfl:
just to avoid any misunderstanding. I didn't ask for fixing the failing test immediately, I was more interested to hear your opinion about whether it is worth to disable optimisation or rather enable optimisation but maybe add a comment to the build script that there is a failing test that will need further investigation. If you are OK with the exceptions test failing I think we should rather keep optimisation enabled and add a comment that the exceptions test needs investigation?

If the other changes look good to you, I'll squash the commits and get it ready for integration.

In any case: once this is integrated there is a baseline where more improvements can happen on top!

dcstes:
I'm OK with enabling -O2  (that is simply using the CFLAGS as set by default when gcc 10 is used).  I don't mind the failing test, I just happened to notice that one test.  But it's true that this can or could be fixed in the future if I am able to find out (or if somebody knows and can help me with it).

The package contents could be in /opt/ooce/gnu-smalltalk instead of /opt/ooce/smalltalk.

I think the path /opt/ooce/gnu-smalltalk would then be better given the package name that is chosen in omnios-extra.   

In Ubuntu/Debian they use the name gnu-smalltalk.   In FreeBSD they use the name smalltalk. 

See https://svnweb.freebsd.org/ports/head/lang/smalltalk/

But my feeling is that gnu-smalltalk describes better the dialect.   There are many smalltalk dialects.

What do you think ? Would /opt/ooce/gnu-smalltalk be acceptable ?

Thanks ...

hadfl (approved):

hadfl:
changed `PREFIX` to `/opt/ooce/gnu-smalltalk`

@cstes please let me know if you are happy with how it looks now.

dcstes:
This is ok for me. 

I tested some simple Smalltalk classes, written by myself,  with this version of GNU Smalltalk, on OmniOS CE, and that works for me, so the fact that some tests are failing (like the exception or Iconv test) is by itself not a big problem to me.

I do not know the authors of GNU Smalltalk, I am only a user of the package.

citrus-it (approved):

hadfl:
Thank you David for the contribution and your patience. The packages will be published soon to the extra repositories.

PR #813, opened 2021-02-21
add rmvb support to minidlna

armgong:
the rmvb support patch is   from openwrt , it very  simple  ,could merge it to minidlna? 
the openwrt patch locate at 
https://github.com/openwrt/packages/blob/master/multimedia/minidlna/patches/005-added-support-RMVB.patch


hadfl (approved):

armgong:
already merge by @citrus-it , so close this

citrus-it:
This has been integrated in a4aaacb8ff92398d39f9447b49fcffa42dee119e
Thank you very much for the contribution - the updated `minidlna` package will be available in the package repository shortly.

PR #825, opened 2021-03-06
merge protobuf build into mosh build

pbdigital-dot-org:


hadfl (dismissed):

citrus-it (commented):
Have you thought about whether we should publish an obsolete package for protobuf to uninstall it from people's systems?

pbdigital-dot-org:
> Have you thought about whether we should publish an obsolete package for protobuf to uninstall it from people's systems?

that has been added now

hadfl (approved):

citrus-it (approved):

PR #849, opened 2021-03-31
ffmpeg enable ssl

cbinner:
Currently ffmpeg doesn't support ssl. I tripped over this when trying to capture a HLS livestream from a https server using youtube-dl.
This commit has `--enable-openssl`, some linux distros seem to prefer gnutls instead but that isn't packaged for omnios.

### DIFF
--- a/build/ffmpeg/build.sh
+++ b/build/ffmpeg/build.sh
@@ -50,7 +50,7 @@ CONFIGURE_OPTS="
     --enable-gpl
     --enable-libx264
     --enable-libx265
-    --enable-openssl
+	--enable-gnutls
 "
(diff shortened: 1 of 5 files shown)


oetiker (dismissed):

citrus-it:
I'm not a lawyer, but from what I've read it seems that we are not permitted to distribute `ffmpeg` linked to `openssl` if we have GPL modules such as x264/265 in the mix.

cbinner:
Sorry I didn't catch this, I'm still figuring out how to build packages. You're right, appently it is impossible with the current openssl version - the next major version will be Apache-licensed and GPL compatible.
So currently SSL support would have to be done with gnutls (it is in pkgsrc but not packaged in OmniOS).

PR #851, opened 2021-04-01
Ffmpeg ssl

cbinner:
Could you take a look at these changes instead of the openssl backend for ffmpeg?

It adds gnutls and its dependancy nettle, I don’t know what the OmniOS policy on pulling in additional crypto libraries is.
I’m not sure about gnutls’ build options, these seem to require the minimal number of external libraries.

There were linker messages suggesting `-zignore` so additional cleanup may be needed.

### DIFF
--- a/build/ffmpeg/build.sh
+++ b/build/ffmpeg/build.sh
@@ -50,7 +50,7 @@ CONFIGURE_OPTS="
     --enable-gpl
     --enable-libx264
     --enable-libx265
-    --enable-openssl
+	--enable-gnutls
 "
 [ $RELVER -ge 151036 ] && CONFIGURE_OPTS+=" --enable-libdav1d"
 CONFIGURE_OPTS_32="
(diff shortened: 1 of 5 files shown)


hadfl:
Hi @cbinner, thanks for this I think it would be good to get GNUtls into the extra package repository to use for cases like ffmpeg.
I think we should add nettle as a full shared library in the usual locations, rather than just building as a static library and when I tested that I found some other changes needed in our unbound package. I hope you don't mind but I've gone ahead and drafted a new branch with these changes:
https://github.com/omniosorg/omnios-extra/compare/master...hadfl:gnutls?expand=1

Please have a look and let us know what you think. If it is ok for you I'd like to add you as the author of these commits.

cbinner:
Thanks, looks good!

hadfl:
closing in favour of https://github.com/omniosorg/omnios-extra/pull/852

PR #880, opened 2021-05-04
add ncdu 1.15.1

jclulow:


oetiker (changes requested):
an entryu in https://github.com/omniosorg/omnios-extra/blob/master/doc/packages.md is missing

build/ncdu/build.sh
```diff
@@ -0,0 +1,43 @@
...
+SUMMARY="$PROG - NCurses Disk Usage"
+DESC='Disk usage analyzer with an ncurses interface'
+
+set_arch 64
+
+set_mirror 'https://dev.yorhel.nl/download/'
+set_checksum sha256 \
+    'b02ddc4dbf1db139cc6fbbe2f54a282770380f0ca5c17089855eab52a9ea3fb0'
+
+CFLAGS+=' -I/usr/include/ncurses '
```
citrus-it, line 33: Does this work with CPPFLAGS instead?

build/ncdu/build.sh
```diff
@@ -0,0 +1,43 @@
...
+VER=1.15.1
+PKG=ooce/util/ncdu
+SUMMARY="$PROG - NCurses Disk Usage"
+DESC='Disk usage analyzer with an ncurses interface'
+
+set_arch 64
+
+set_mirror 'https://dev.yorhel.nl/download/'
+set_checksum sha256 \
+    'b02ddc4dbf1db139cc6fbbe2f54a282770380f0ca5c17089855eab52a9ea3fb0'
```
citrus-it, line 31: I've put the source on the omnios mirror now, so please drop these three lines

build/ncdu/build.sh
citrus-it, line 33: Does this work with CPPFLAGS instead?

build/ncdu/local.mog
```diff
@@ -0,0 +1,18 @@
...
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+# }}}
+#
+# Copyright 2021 Oxide Computer Company
+
+license COPYING license=MIT
+
+# vim:ts=4:sw=4:et:fdm=marker
```
citrus-it, line 18: We generally do not use the fold markers of the vim modeline in the .mog files, just a trailing blank line.

citrus-it (changes requested):
Thanks Josh, that looks like a useful utility. Just a few nits, and please can you add the package to `doc/packages.md` so that our package update monitoring plugins can keep an eye on updates? 
Setting the URL there to https://dev.yorhel.nl/download should be enough for it to work. Please set the maintainer field to Oxide if they/you will be looking after it.

Can you also please add a line to `doc/baseline`?

build/ncdu/build.sh
```diff
@@ -0,0 +1,43 @@
...
+
+CFLAGS+=' -I/usr/include/ncurses '
+
+init
+download_source "" "$PROG" "$VER"
+prep_build
+build
+strip_install
```
citrus-it, line 39: Since CTF conversion seems to work on all OmniOS releases, can you replace these two lines?

```suggestion
build -ctf
```

jclulow:
With my apologies for it taking so long, I think I've addressed all the feedback!  I've confirmed that the package still builds.  Let me know if you need anything else.

hadfl (approved):

PR #892, opened 2021-05-14
Build upstream mod_md instead of bundle copy

sjorge:
Apache syncs from upstream every so often, unfortunately it seems it is currently a bit behind.

The bundled mod_md works fine with boulder based acme registries (aka letsencrypt) but segfaults when using others e.g. pebble based or something like smallstalk's step-ca. (Guess what I am using for my non internet exposed hosts)

This builds mod_md seperately from apache, as a small bonus this drops the web/curl and ooce/library/jansson dependencies again for the main apache-24 package, as only mod_md needs them.

I'm not sure happy with this as I'd prefer to use the bundled one, as that is one package less to track/update. But this is better than having httpd die with a segfault.

### DIFF
--- a/build/mod_md/build.sh
+++ b/build/mod_md/build.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+
+# Copyright 2021 OmniOS Community Edition (OmniOSce) Association.
+
+. ../../lib/functions.sh
+
+PROG=mod_md
+VER=2.4.0
+# Hard-coded here for now. If we ship more than one apache version, this will
+# need restructuring.
+PKG=ooce/server/apache-24/modules/md
+SUMMARY="$PROG Let's Encrypt (ACME) support for Apache httpd"
+DESC="$SUMMARY"
+
+APACHEVER=2.4
+sAPACHEVER=${APACHEVER//./}
+
+RUN_DEPENDS_IPS+=" ooce/server/apache-$sAPACHEVER"
+
+set_arch 64
+
+set_mirror 'https://github.com/icing/mod_md/releases/download/'
+set_checksum sha256 \
+    '1710ffe0931a396f155c922d6873e56e3d4947c46f39094a428a766672f1ef91'
+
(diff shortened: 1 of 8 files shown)


build/mod_md/build.sh
```diff
@@ -0,0 +1,63 @@
...
+
+XFORM_ARGS="
+    -DPREFIX=${PREFIX#/}
+    -DOPREFIX=${OPREFIX#/}
+    -DPROG=$PROG
+"
+
+export PATH+=":$PREFIX/bin"
+LDFLAGS64+=" -L$OPREFIX/lib/$ISAPART64 -R$OPREFIX/lib/$ISAPART64"
+CFLAGS64+=" -D_XOPEN_SOURCE=700"
```
sjorge, line 49: This is needed, otherwise apr complains about unknown PATH_MAX, mod_md itself also complains about implicitly defined kill().

build/mod_md/files/httpd-md.conf
```diff
@@ -0,0 +1,8 @@
+LoadModule	watchdog_module libexec/mod_watchdog.so
```
sjorge, line 1: Used the opportunity to add a example config.

build/mod_md/patches/arc4random_buf.patch
```diff
@@ -0,0 +1,22 @@
+Patch origin: in-house
```
sjorge, line 1: I'd like to see this patch go, but after 3 hours of struggling with this I can't get past the compile error.

We have arc4random_buf in stdlib.h which is included in md_crypt.c, but for some reason the compiler is not seeing it. With the check removed from configure mod_crypto.c will fall back to a custom seed_RAND function.

sjorge:
I've left some remarks about why I am not super happy with this, but if this is acceptable I'll polish this a bit and push the  doc/packages.md and doc/baseline which I forgot.

citrus-it:
> I'm not sure happy with this as I'd prefer to use the bundled one, as that is one package less to track/update. But this is better than having httpd die with a segfault.

Can you not just add the patch from upstream to fix the crash you're experiencing? Once the next apache comes out, hopefully it could be dropped again.

sjorge:
> > I'm not sure happy with this as I'd prefer to use the bundled one, as that is one package less to track/update. But this is better than having httpd die with a segfault.
> 
> Can you not just add the patch from upstream to fix the crash you're experiencing? Once the next apache comes out, hopefully it could be dropped again.

If I knew which commit exactly it was, there were a lot.

sjorge:
Tbh, I was kind of hoping you say you'd like to stick with the bundled one. I could abandon the private acme endpoint all together though. 

Not sure many people would use it for !letsencrypt ... also probably why so few people care about it. As everyone is using letsencrypt.

sjorge:
Lets just close this, and stick to the bundle one, hopefully a 2.4.48 eventually get release that has the same fixes.

PR #895, opened 2021-05-17
Prevent mod_md segfault with non letsencrypt acme

sjorge:
We pull in these commits from upstream that sync up mod_md:
c0cff48bccf6607333a6841f723660c0083929f4
7da5270ffe75101b19e1ca31b287c5afa2d44e8a

Both of these patches should be able to be
dropped once 2.4.48 is released.

We also need to run autoconf, as configure is not shipped in the trunk
repo, we need to regen it to get build mod_md after the changes.

### DIFF
--- a/build/apache/build-24.sh
+++ b/build/apache/build-24.sh
@@ -72,7 +72,7 @@ CONFIGURE_OPTS="
 init
 download_source $PROG httpd $VER
 patch_source
-prep_build
+prep_build autoconf -autoreconf
 build -ctf
 xform files/apache-template.xml > $TMPDIR/$PROG-$sMAJVER.xml
 install_smf ooce $PROG-$sMAJVER.xml
--- a/build/apache/patches-24/series
+++ b/build/apache/patches-24/series
@@ -1,3 +1,5 @@
 layout.patch
 priv_drop.patch
 httpd.conf.patch
+upstream-md-sync-part1.patch
+upstream-md-sync-part2.patch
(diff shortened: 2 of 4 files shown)


sjorge:
This is the 2nd attempt, as Andy suggestion, to pull in the httpd upstream commits earlier.

The mentioned 2 commits were cherry-picked from the 2.4.x branch on github, so those are scheduled to be in 2.4.48. So we can drop them on the next release.

Given that `./configure` is not shipped in the repo, it needs to be regenerated from `./configure.in` by running autoconf to pick up the mod_md changes.

oetiker (approved):

citrus-it:
sjorge - it looks like Apache httpd 2.4.48 is a couple of days away from release, along with "substantial" changes to mod_md

https://lists.apache.org/list.html?dev@httpd.apache.org

Since it's so close, can we wait for that rather than backporting?

sjorge:
Sounds reasonable, between the last to release was a super long time, if there is a new one rather quick waiting seems the way to go.

Although there are already some  veto’s wrt libressl builds. 

Let’s keep this open until for a bit, say… let’s wait until July first to see if they do a release? 

Or is that to short of a time to wait it out?

~ sjorge

sjorge:
[12:32:52] sjorge: .46 was in august 2020, .47 April 2021
[12:33:30] sjorge: waiting a month or so for a .48 seems reasonable, waiting another 8 months less so

citrus-it:
Superseded by #901

PR #917, opened 2021-07-01
Update Virtualbox to support illumos 13915

citrus-it:
Tested that this updated virtualbox package works on r151038 both before and after illumos 13915.

```
theeo% uname -a
SunOS theeo 5.11 omnios-r151038-a4c5b5181d i86pc i386 i86pc   <- upcoming r151038k

theeo% vagrant up
Bringing machine 'default' up with 'virtualbox' provider...

theeo% vagrant ssh
Linux debian-10 4.19.0-9-amd64 #1 SMP Debian 4.19.118-2+deb10u1 (2020-06-07) x86_64

Last login: Sat May  1 20:07:19 2021 from 10.0.2.2
vagrant@debian-10:~$
```

### DIFF
--- a/build/virtualbox/patches/build-in-zone.patch
+++ b/build/virtualbox/patches/build-in-zone.patch
@@ -3,7 +3,7 @@ Patch originally taken from OpenIndiana, author is Alexander Pyhalov
 diff -wpruN '--exclude=*.orig' a~/Config.kmk a/Config.kmk
 --- a~/Config.kmk	1970-01-01 00:00:00
 +++ a/Config.kmk	1970-01-01 00:00:00
-@@ -2749,8 +2749,6 @@ $(PATH_OUT)/DynamicConfig.kmk: \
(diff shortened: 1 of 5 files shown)


danmcd (commented):
Are you sure installctx() is being called without kernel thread preemption already enabled?  Patches-of-patches makes this difficult to answer.  :)

citrus-it:
> Are you sure installctx() is being called without kernel thread preemption already enabled? Patches-of-patches makes this difficult to answer. :)

I am, there's a comment in the code directly above the call which says:

```
    /*
     * installctx() allocates memory and thus cannot be used in RTThreadCtxHookRegister() which can be used
     * with preemption disabled. We allocate the context-hooks here and use 'fEnabled' to determine if we can
     * invoke the consumer's hook or not.
     */
```

danmcd (approved):
Thank you, my bad (the first line is cut-off in the patch).

oetiker (approved):

hadfl (approved):

PR #938, opened 2021-08-04
add mosquitto

sjorge:
Mosquitto is a MQTT Broker

It is complaining about an unknown license type, but I also noticed there are a few more of those when you grep for 'license=' and compare them against the pattern file.

Whats the play here? Do I update the pattern file or leave as is?

### DIFF
--- a/build/mosquitto/patches/mux_epoll_c.patch
+++ b/build/mosquitto/patches/mux_epoll_c.patch
@@ -1,0 +1,10 @@
+--- a/src/mux_epoll.c	Wed Aug  4 12:43:36 2021
++++ b/src/mux_epoll.c	Wed Aug  4 12:44:11 2021
+@@ -28,6 +28,7 @@
+ #include <assert.h>
+ #ifndef WIN32
+ #ifdef WITH_EPOLL
++#include <signal.h>
+ #include <sys/epoll.h>
+ #define MAX_EVENTS 1000
+ #endif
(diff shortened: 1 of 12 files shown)


build/mosquitto/patches/mux_epoll_c.patch
```diff
@@ -0,0 +1,10 @@
+--- a/src/mux_epoll.c	Wed Aug  4 12:43:36 2021
++++ b/src/mux_epoll.c	Wed Aug  4 12:44:11 2021
+@@ -28,6 +28,7 @@
+ #include <assert.h>
+ #ifndef WIN32
+ #ifdef WITH_EPOLL
++#include <signal.h>
```
oetiker, line 7: if you want to upstream this the signal include should be conditional ...

build/mosquitto/patches/mux_epoll_c.patch
```diff
@@ -0,0 +1,10 @@
...
+ #include <sys/epoll.h>
+ #define MAX_EVENTS 1000
+ #endif
```
oetiker, line 10: you seem to open two ifs but have only one endif ?

also can the assert.h be loaded in any environment ?

sjorge: I’m only inserting signal.h above says/poll.h, the other stuff is already in the file.

git does a poor job of showing patches.

oetiker: 🙈 sorry

hadfl (approved):

citrus-it (approved):

PR #1056, opened 2022-02-18
postgresql use ISM by default and add DISM support.

mvwieringen:
This adds support for using Intimate Shared Memory using the sysv SHM
methods and Dynamic Intimate Shared Memory when available by checking
for the SHM_PAGEABLE flag in <sys/shm.h>. This makes it possible to have
ISM/DISM without needing to set the shared_memory_type and
dynamic_shared_memory_type to sysv in postgresql.conf. The setting of
shared_memory_type is something that seems to origin from Joyent for
Manta and is available in upstream versions 12 and later so we only
patch those versions for now. Dropped --with-xslt configure options
as all versions issue an unrecognized option for that configure option.

### DIFF
--- a/build/postgresql/build-14.sh
+++ b/build/postgresql/build-14.sh
@@ -57,8 +57,9 @@ CONFIGURE_OPTS="
     --localstatedir=$VARPATH
     --enable-thread-safety
     --with-openssl
+    --with-lz4
     --with-libxml
-    --with-xslt
+    --with-libxslt
     --with-readline
     --without-systemd
     --with-system-tzdata=/usr/share/lib/zoneinfo
--- a/build/postgresql/build-14.sh
+++ b/build/postgresql/build-14.sh
@@ -76,9 +76,11 @@ MAKE_INSTALL_TARGET=install-world
 
 build_manifests() {
     manifest_start $TMPDIR/manifest.client
-    manifest_add_dir $PREFIX/include libpq
+    manifest_add_dir $PREFIX/include '.*'
     manifest_add_dir $PREFIX/lib/pkgconfig
     manifest_add_dir $PREFIX/lib/$ISAPART64/pkgconfig
+    manifest_add_dir $PREFIX/lib/pgxs '.*'
+    manifest_add_dir $PREFIX/lib/$ISAPART64/pgxs '.*'
     manifest_add $PREFIX/lib '.*lib(pq\.|ecpg|pgtypes|pgcommon|pgport).*'
     manifest_add $PREFIX/bin '.*pg_config' psql ecpg
     manifest_add $PREFIX/share/man/man1 pg_config.1 psql.1 ecpg.1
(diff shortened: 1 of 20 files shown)


build/postgresql/patches-14/dism.patch
```diff
@@ -0,0 +1,89 @@
...
+ extern void *UsedShmemSegAddr;
+ 
+-#if !defined(WIN32) && !defined(EXEC_BACKEND)
++#if !defined(WIN32) && !defined(EXEC_BACKEND) && !defined(SHM_SHARE_MMU)
+ #define DEFAULT_SHARED_MEMORY_TYPE SHMEM_TYPE_MMAP
+-#elif !defined(WIN32)
++#elif !defined(WIN32) || defined(SHM_SHARE_MMU)
+ #define DEFAULT_SHARED_MEMORY_TYPE SHMEM_TYPE_SYSV
+ #else
+ #define DEFAULT_SHARED_MEMORY_TYPE SHMEM_TYPE_WINDOWS
```
citrus-it, line 89: It took me a minute to parse this, it might be better to patch it like:

```diff
@@ -70,7 +70,9 @@
 #endif
 extern void *UsedShmemSegAddr;

+#if defined(SHM_SHARE_MMU)
+#define DEFAULT_SHARED_MEMORY_TYPE SHMEM_TYPE_SYSV
-#if !defined(WIN32) && !defined(EXEC_BACKEND)
+#elif !defined(WIN32) && !defined(EXEC_BACKEND)
 #define DEFAULT_SHARED_MEMORY_TYPE SHMEM_TYPE_MMAP
 #elif !defined(WIN32)
 #define DEFAULT_SHARED_MEMORY_TYPE SHMEM_TYPE_SYSV
```

Happy either way, it's obviously doing the right thing.

citrus-it (approved):

build/postgresql/patches-14/dism.patch
mvwieringen: Yeah I already did the other for dynamic shared memory to be somewhat better readable. The patch however looks a bit different then what you pasted :-) diff does the grouping just a bit differently it seems. I tested with -xx on all 3 build scripts and the new dism.patch still applies to all of them now.

build/postgresql/build-14.sh
```diff
@@ -50,22 +50,25 @@ XFORM_ARGS="
 
 CFLAGS+=" -O3"
 CPPFLAGS+=" -DWAIT_USE_POLL"
+LDFLAGS64+=" -R /opt/ooce/llvm-$DEFAULT_CLANG_VER/lib"
```
citrus-it, line 53:
```suggestion
LDFLAGS64+=" -R $OPREFIX/llvm-$DEFAULT_CLANG_VER/lib"
```

citrus-it (changes requested):

citrus-it (dismissed):

oetiker (dismissed):

hadfl (approved):

citrus-it (approved):

PR #1057, opened 2022-02-20
add citus postgresql extension.

mvwieringen:
This is an extension that makes postgresql an distributed database its a plugin that is loaded. I looked at OpenIndiana as to what they provide and picked up this idea. Initial testing shows it loads but the packaging needs some help.


mvwieringen:
will put it in the other PR.

PR #1060, opened 2022-02-24
enable JIT/llvm support for postgresql14.

mvwieringen:
This statically links the llvm stuff into /opt/ooce/pgsql-14/lib/amd64/llvmjit.so which leads to
it being around 50 Mb e.g. pkg.csize=18194824 pkg.size=50894672

I first tried dynamic linking but the we need  /opt/ooce/llvm-13/lib/libLLVM-13.so which is 116967096 bytes.

So statically linking saves us about 56% thanks hadfl for the hint.

### DIFF
--- a/build/postgresql/build-14.sh
+++ b/build/postgresql/build-14.sh
@@ -25,6 +25,12 @@ DESC="The World's Most Advanced Open Source Relational Database"
 
 SKIP_LICENCES=postgresql
 
+# We want to populate the clang-related environment variables
+# and set PATH to point to the correct llvm/clang version for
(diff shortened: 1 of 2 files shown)


mvwieringen:
Tested using the tips from

https://severalnines.com/database-blog/overview-just-time-compilation-jit-postgresql

truss-ed the postgresql14 binaries and it load the llvmjit.so file correctly and maps it into the postgres binary.

Output from explain commands also shows that JIT is used analog to what is shown in the linked blog.

oetiker (approved):

citrus-it (approved):

PR #1064, opened 2022-02-25
Add mysql(mariadb) foreign data wrapper for Postgresql.

mvwieringen:
Tested by loading the extension into pg14. pldd shows:

/opt/ooce/pgsql-14/lib/amd64/mysql_fdw.so
/opt/ooce/mariadb-10.4/lib/amd64/libmariadb.so.3

Some fixes along the way. mysql_fdw uses some prebuild Makefile which need the USE_PGXS=1 flag
to be usable that works for make and make install but not for make clean and make distclean so extended
the make_clean() function in lib/functions.sh

Also need to call clean_up in the version loop. For citus it kind of works for mysql_fdw not so much and you
end up with multiple versions being installed into the proto root which is being packaged and as such the second
build has both pg13 and pg14 files.

The patches needed for mysql_fdw are partly from OpenIndiana but it seems that the current release 2.7.0 is
not compatible with pg14.2 so needed an other patch. The patch has some comments as to why its needed.


lib/functions.sh
```diff
@@ -2164,9 +2164,11 @@ make_clean() {
         CLEAN_SOURCE=
         return
     fi
+
+    eval set -- $MAKE_CLEAN_ARGS_WS
     logmsg "--- make (dist)clean"
     (
-        $MAKE distclean || $MAKE clean
+        $MAKE $MAKE_CLEAN_ARGS distclean || $MAKE $MAKE_CLEAN_ARGS clean
```
citrus-it, line 2171: You set the args list from the whitespace variant (`MAKE_CLEAN_ARGS_WS`) but then don't use it. Add `$@` like this to use the result of that:

```suggestion
        $MAKE $MAKE_CLEAN_ARGS "$@" distclean \
            || $MAKE $MAKE_CLEAN_ARGS "$@" clean
```

Can you also add these variables to `lib/config.sh`, where the other MAKE_ARGS ones are?

citrus-it (changes requested):

hadfl (dismissed):

citrus-it (dismissed):

citrus-it (approved):

PR #1071, opened 2022-03-08
dnsmasq: add package (ver 2.86)

bahamas10:
Hello! This PR adds the `dnsmasq` package.

Possible Concerns
-------------------

This is my first PR for omnios-extra so it's possible I got things wrong or incorrect here.  A couple of possible concerns I had are:

### 1. SMF category

I currently have this under the `network` category, but should it be under `ooce`? I'm not sure what the preference here is for that.  FWIW I based my implementation for this service off of the `mosquitto` package which uses the `network` category.

### 2. User ID and Group ID conflicts

I basically arbitrarily picked a UID and GID that didn't seem to be in use (i used a simple `grep` over the `build/` directory) - is there a better way to pick a UID and GID that won't possibly conflict?

Notes
------

The patches were directly copied from Joyent/pkgsrc for SmartOS (https://github.com/joyent/pkgsrc/tree/trunk/net/dnsmasq/patches)

Testing
--------

I installed the package manually and verified the compilation worked as expected by running the binary with `-v` (version):

```
dave - test-zone sunos ~/dev/omnios-extra (git:dnsmasq) $ pkg info dnsmasq
             Name: ooce/network/dnsmasq
          Summary: Lightweight, easy to configure DNS forwarder
      Description: dnsmasq is a lightweight, easy to configure DNS forwarder,
                   designed to provide DNS (and optionally DHCP and TFTP)
                   services to a small-scale network.
            State: Installed (Manually installed)
        Publisher: local.omnios
          Version: 2.86
           Branch: 151040.0
   Packaging Date: March  8, 2022 at 09:49:00 AM
Last Install Time: March  8, 2022 at 09:49:40 AM
             Size: 605.31 kB
             FMRI: pkg://local.omnios/ooce/network/dnsmasq@2.86-151040.0:20220308T094900Z
```

```
dave - test-zone sunos ~/dev/omnios-extra (git:dnsmasq) $ /opt/ooce/sbin/dnsmasq -v
Dnsmasq version   Copyright (c) 2000-2021 Simon Kelley
Compile time options: IPv6 GNU-getopt no-DBus no-UBus no-i18n no-IDN DHCP DHCPv6 no-Lua TFTP no-conntrack no-ipset auth no-cryptohash no-DNSSEC loop-detect no-inotify dumpfile

This software comes with ABSOLUTELY NO WARRANTY.
Dnsmasq is free software, and you are welcome to redistribute it
under the terms of the GNU General Public License, version 2 or 3.
```

I then used a small config file to test that the service works as expected:

```
$ sudo cp dnsmasq.conf /etc/opt/ooce/dnsmasq/dnsmasq.conf
$ sudo svcadm enable dnsmasq
$ dig foo.example.com @localhost +short A
10.0.1.50
$ dig bar.example.com @localhost +short A
10.0.1.51
$ dig google.com @localhost +short A
142.251.40.142
```

```
$ tail /var/svc/log/network-dnsmasq\:default.log
[ Mar  8 10:04:09 Method "start" exited with status 0. ]
Mar  8 10:04:16 dnsmasq[12300]: query[A] foo.example.com from 127.0.0.1
Mar  8 10:04:16 dnsmasq[12300]: config foo.example.com is 10.0.1.50
Mar  8 10:04:17 dnsmasq[12300]: query[A] bar.example.com from 127.0.0.1
Mar  8 10:04:17 dnsmasq[12300]: config bar.example.com is 10.0.1.51
Mar  8 10:04:18 dnsmasq[12300]: query[A] google.com from 127.0.0.1
Mar  8 10:04:18 dnsmasq[12300]: forwarded google.com to 1.1.1.1
Mar  8 10:04:19 dnsmasq[12300]: reply google.com is 142.250.65.206
```

---

`dnsmasq.conf`

``` ini
port=53
listen-address=0.0.0.0
bind-interfaces

domain-needed
proxy-dnssec
filterwin2k
no-resolv
no-poll
no-hosts
neg-ttl=3600
local-ttl=300

# log queries to stderr
log-queries
log-facility=-

# Custom
domain=example.com
local=/example.com/

# Custom DNS
address=/foo.example.com/10.0.1.50
address=/bar.example.com/10.0.1.51

# PTR records
ptr-record=50.1.0.10.in-addr.arpa,"foo.example.com"
ptr-record=51.1.0.10.in-addr.arpa,"bar.example.com"

# Forward everything else to cloudflare
server=1.1.1.1
```

### DIFF
--- a/build/dnsmasq/build.sh
+++ b/build/dnsmasq/build.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+
+# Copyright 2022 OmniOS Community Edition (OmniOSce) Association.
+
+. ../../lib/build.sh
+
+PROG=dnsmasq
+VER=2.86
+PKG=ooce/network/dnsmasq
+SUMMARY="Lightweight, easy to configure DNS forwarder"
+DESC="dnsmasq is a lightweight, easy to configure DNS forwarder, designed to "
+DESC+="provide DNS (and optionally DHCP and TFTP) services to a small-scale network."
+
+set_arch 64
+set_mirror 'https://thekelleys.org.uk/'
+set_checksum sha256 '28d52cfc9e2004ac4f85274f52b32e1647b4dbc9761b82e7de1e41c49907eb08'
+
+BASEDIR=$PREFIX/$PROG
+CONFFILE=/etc$BASEDIR/$PROG.conf
+EXECFILE=$PREFIX/sbin/dnsmasq
+
+copy_sample_config() {
+    local relative_conffile=${CONFFILE#/}
+    local dest_confdir=$DESTDIR/${relative_conffile%/*}
+
+    logmsg "-- copying sample config"
+    logcmd mkdir -p "$dest_confdir" || logerr "mkdir failed"
+    logcmd cp $TMPDIR/$BUILDDIR/$PROG.conf.example $DESTDIR/$relative_conffile \
+        || logerr "copying configs failed"
+}
+
+# No configure
+configure64() { :; }
+
+XFORM_ARGS="
+    -DPREFIX=${PREFIX#/}
+    -DBASEDIR=${BASEDIR#/}
+    -DEXECFILE=$EXECFILE
+    -DCONFFILE=$CONFFILE
+    -DUSER=$PROG
+    -DGROUP=$PROG
+    -DPROG=$PROG
+"
+
+MAKE_ARGS_WS="
+    CC=$CC
+    CFLAGS=\"-DNO_IPSET $CFLAGS $CFLAGS64 $CTF_CFLAGS\"
+    LDFLAGS=\"$LDFLAGS $LDFLAGS64\"
+    PREFIX=$PREFIX
+    MANDIR=$PREFIX/share/man
+    sunos_libs=\"-lnsl -lsocket\"
+"
+
+MAKE_INSTALL_ARGS_WS="
+    PREFIX=$PREFIX
+    MANDIR=$PREFIX/share/man
+"
+
+init
+download_source "$PROG" "$PROG" "$VER"
+patch_source
+prep_build
+build
+copy_sample_config
+xform files/$PROG.xml > $TMPDIR/$PROG.xml
+install_smf network $PROG.xml
+make_package
+clean_up
+
+# Vim hints
+# vim:ts=4:sw=4:et:fdm=marker
--- a/build/dnsmasq/build.sh
+++ b/build/dnsmasq/build.sh
@@ -24,8 +24,6 @@ DESC="dnsmasq is a lightweight, easy to configure DNS forwarder, designed to "
 DESC+="provide DNS (and optionally DHCP and TFTP) services to a small-scale network."
 
 set_arch 64
-set_mirror 'https://thekelleys.org.uk/'
-set_checksum sha256 '28d52cfc9e2004ac4f85274f52b32e1647b4dbc9761b82e7de1e41c49907eb08'
 
 BASEDIR=$PREFIX/$PROG
 CONFFILE=/etc$BASEDIR/$PROG.conf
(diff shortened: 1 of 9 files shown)


danmcd:
This looks like an alternative to unbound (it's a DNS forwarder).  I'd use unbound's examples for naming and SMF:

https://pkg.omnios.org/bloody/extra/info/0/ooce%2Fnetwork%2Funbound%401.15.0%2C5.11-151041.0%3A20220212T132554Z

```
dns-zone(0)# svcs -a | grep unbound
online         Feb_01   svc:/network/dns/unbound:default
dns-zone(0)# 
```

danmcd:
I'm glad an extra client-side resolver is coming to OOCE.  Choice is good!

bahamas10:
> This looks like an alternative to unbound (it's a DNS forwarder). I'd use unbound's examples for naming and SMF:
> 
> https://pkg.omnios.org/bloody/extra/info/0/ooce%2Fnetwork%2Funbound%401.15.0%2C5.11-151041.0%3A20220212T132554Z
> 
> ```
> dns-zone(0)# svcs -a | grep unbound
> online         Feb_01   svc:/network/dns/unbound:default
> dns-zone(0)# 
> ```

gotcha, the `svc:/network/dns/dnsmasq:default` FMRI change makes sense, i'll make that change tomorrow and push it.

build/dnsmasq/build.sh
```diff
@@ -0,0 +1,83 @@
...
+    -DEXECFILE=$EXECFILE
+    -DCONFFILE=$CONFFILE
+    -DUSER=$PROG
+    -DGROUP=$PROG
+    -DPROG=$PROG
+"
+
+MAKE_ARGS_WS="
+    CC=$CC
+    CFLAGS=\"-DNO_IPSET $CFLAGS $CFLAGS64 $CTF_CFLAGS\"
```
citrus-it, line 59: This package will be built for a range of different OmniOS versions, and not all of those build packages with CTF, so you only want to conditionally include CTF_CFLAGS here.
The best way to deal with this is to wrap the assignments for `MAKE_ARGS_WS` and `MAKE_INSTALL_ARGS_WS` into the `configure64()` function, and drop `$CTF_CFLAGS`.
For those versions where CTF is the default, `CFLAGS` will be automatically extended with `CTF_CFLAGS` in the `build()` function which is what ends up calling configure64().
https://github.com/omniosorg/omnios-extra/blob/master/build/lz4/build.sh is an example of this.

build/dnsmasq/build.sh
```diff
@@ -0,0 +1,83 @@
...
+"
+
+init
+download_source "$PROG" "$PROG" "$VER"
+patch_source
+prep_build
+build
+copy_sample_config
+xform files/$PROG.xml > $TMPDIR/$PROG.xml
+install_smf network $PROG.xml
```
citrus-it, line 78: Can you please change to this? We're trying to adopt an approach of keeping the 'extra' manifests in a separate subdirectory. The service should still be `svc:/network/dns/dnsmasq:default` - no ooce there.
```suggestion
install_smf ooce $PROG.xml
```

build/dnsmasq/files/dnsmasq.xml
```diff
@@ -0,0 +1,79 @@
...
+<service_bundle type="manifest"
+                name="network:dnsmasq">
+
+    <service name="network/dnsmasq"
+             type="service"
+             version="1">
+
+        <create_default_instance enabled="false" />
+
+        <single_instance />
```
citrus-it, line 25: Drop this line, see https://www.illumos.org/issues/12071

build/dnsmasq/local.mog
```diff
@@ -0,0 +1,20 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+# Copyright 2022 OmniOS Community Edition (OmniOSce) Association.
+
+dir group=$(GROUP) mode=0770 owner=$(USER) path=var/$(BASEDIR)
+
+group groupname=$(GROUP) gid=99
+user ftpuser=false username=$(USER) uid=99 group=$(GROUP) \
```
citrus-it, line 16: UIDs are tracked in the `doc/idlist.md` file in this repository, covering the users that come from illumos-omnios, omnios-build as well as extra.
99 is free, but I'd suggest using 56 since that groups it with the other DNS services
(and please add rows to that file).

citrus-it (changes requested):
Thanks for the contribution, this looks really good. I just have a few comments, and answers to your questions.

Please can you also update the `doc/packages.md` and `doc/baseline` files with this new package?

Thanks

build/dnsmasq/local.mog
```diff
@@ -0,0 +1,20 @@
...
+    gcos-field="dnsmasq - DNS Forwarder" \
+    home-dir=/var/$(BASEDIR) password=NP
+
```
citrus-it, line 19: To stop the configuration file being overwritten on package updates, you will need to set a preserve flag on it. There are various options (See https://man.omnios.org/pkg.7 ) but I usually favour `renamenew` - so something like:

```
<transform file path=etc/ -> set preserve renamenew>
```

You should also configure the service to automatically restart if the binary changes, which is:

```
<transform file path=$(PREFIX)/sbin -> \
    set restart_fmri svc:/network/dns/$(PROG):default >
```

bahamas10:
> Thanks for the contribution, this looks really good. I just have a few comments, and answers to your questions.
> 
> Please can you also update the `doc/packages.md` and `doc/baseline` files with this new package?
> 
> Thanks

Thank you so much for the thorough (and quick!) review.  I've made all of the changes you outlined and pushed them as a new commit.  I've tested again on my local dev box and everything seems to build and work as expected.   Thank you for taking the time to not only suggest changes but explain why (and how it all works).

citrus-it (approved):
Looks good to me, thanks for the changes.
The extra repository is a bit strange in just having a single branch that has to cover all of the releases.

oetiker (approved):

PR #1092, opened 2022-04-09
vaultwarden: add package (ver 1.24.0)

papertigers:
This is a work in progress but I wanted to get more eyes on it sooner than later.

I have a few questions, so I plan to comment directly on this PR to ask them.

### DIFF
--- a/build/vaultwarden/build.sh
+++ b/build/vaultwarden/build.sh
@@ -1,0 +1,129 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+
+# Copyright 2022 OmniOS Community Edition (OmniOSce) Association.
+
+. ../../lib/build.sh
+
+PROG=vaultwarden
+VER=1.24.0
+PKG=ooce/application/vaultwarden
+SUMMARY="Bitwarden compatible server"
+DESC="Unofficial Bitwarden compatible server written in Rust, formerly known "
+DESC+="as bitwarden_rs"
+
+DANIGARCIA=$GITHUB/dani-garcia
+WEBVAULTVER=2.27.0
+
+# The next vaultwarden version requires only the stable 1.25.0 rust version
+# until then, use nightly.
+RUST=rust-nightly-x86_64-unknown-illumos
+RUSTVER=2022-01-23
+
+set_arch 64
+
+BASEDIR=$PREFIX/$PROG
+CONFFILE=/etc$BASEDIR/env.template
+WEBVAULTDIR=/var$BASEDIR/web-vault
+EXECFILE=$PREFIX/bin/$PROG
+
+BMI_EXPECTED=1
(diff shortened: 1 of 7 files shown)


build/vaultwarden/build.sh
```diff
@@ -0,0 +1,130 @@
...
+get_webvault() {
+    local danigarcia=$GITHUB/dani-garcia
+    local prog_repo=bw_web_builds
+    local prog=web-vault
+    local relative_webvaultdir=${WEBVAULTDIR#/}
+    local dest_webvaultdir=$DESTDIR/${relative_webvaultdir%/*}
+
+    note -n "Pulling v$WEBVAULTVER prebuilt $prog"
+
+    save_variables SKIP_CHECKSUM
```
papertigers, line 77: Is it a good idea to do this save/restore variable dance in this function?

build/vaultwarden/local.mog
```diff
@@ -0,0 +1,26 @@
...
+
+dir group=$(GROUP) mode=0770 owner=$(USER) path=var/$(BASEDIR)
+
+group groupname=$(GROUP) gid=58
+user ftpuser=false username=$(USER) uid=58 group=$(GROUP) \
+    gcos-field="vaultwarden - Password Manager" \
+    home-dir=/var/$(BASEDIR) passwd=NP
+
+# XXX what goes here?
+#<transform file path=etc/ -> set preserve renamenew>
```
papertigers, line 21: Not quite sure what to put here yet as vaultwarden only cares about the files under /var/ooce

papertigers:
Additionally pkglint gives me the following, and I am not sure where my issue is.  If someone could point me in the right direction that would be great!

```
Lint engine setup...
Starting lint run...

ERROR pkglint.dupaction007        dir action for var/opt/ooce/vaultwarden is reference-counted but has different attributes across 1 duplicates: group: bin -> ooce/server/vaultwarden group: vaultwarden -> ooce/serv
er/vaultwarden mode: 0755 -> ooce/server/vaultwarden mode: 0770 -> ooce/server/vaultwarden owner: root -> ooce/server/vaultwarden owner: vaultwarden -> ooce/server/vaultwarden
```

papertigers:
Also note I have not directly tested the pkg in a test zone yet.

build/vaultwarden/build.sh
citrus-it: It can probably be dropped. We can just set `SKIP_CHECKSUM=1` for the download call itself.
Since the web stuff is downloaded last, I think overwriting the variables is fine, and we just need to check that the `info.source-url` entries in the final manifest look ok.

build/vaultwarden/local.mog
citrus-it: I'd move the config file to `/etc` and symlink it from `/var`

citrus-it:
> Additionally pkglint gives me the following, and I am not sure where my issue is. If someone could point me in the right direction that would be great!
> 
> ```
> Lint engine setup...
> Starting lint run...
> 
> ERROR pkglint.dupaction007        dir action for var/opt/ooce/vaultwarden is reference-counted but has different attributes across 1 duplicates: group: bin -> ooce/server/vaultwarden group: vaultwarden -> ooce/serv
> er/vaultwarden mode: 0755 -> ooce/server/vaultwarden mode: 0770 -> ooce/server/vaultwarden owner: root -> ooce/server/vaultwarden owner: vaultwarden -> ooce/server/vaultwarden
> ```

This is because the web-vault installation creates the `var/opt/ooce/vaultwarden` directory, then you create it again in the local.mog.

citrus-it (dismissed):

hadfl (approved):

PR #1095, opened 2022-04-11
Add ansible package

citrus-it:
This may still warrant some slimming down.

```
PKG                                      FLAGS    FILES        SIZE        COMP
ooce/system/management/ansible                    28294  315.69 MiB   78.17 MiB
```

I've gone for the `ansible` package which bundles a lot of plugins


oetiker (approved):
nice

PR #1129, opened 2022-07-04
Add rlwrap

r7st:
Adding rlwrap package. 

From README:
"rlwrap is a 'readline wrapper', a small utility that uses the GNU Readline library to allow the editing of keyboard input for any command."

https://github.com/hanslub42/rlwrap

### DIFF
--- a/build/rlwrap/build.sh
+++ b/build/rlwrap/build.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+
+# Copyright 2022 r7st r7st.guru@gmail.com
+
+. ../../lib/build.sh
+
+PROG=rlwrap
+VER=0.45.2
+PKG=ooce/util/rlwrap
+SUMMARY="readline wrapper"
+DESC="rlwrap is a 'readline wrapper', a small utility that uses the GNU "
+DESC+="Readline library to allow the editing of keyboard input for any command."
+
+set_arch 64
+
+init
+download_source $PROG $PROG $VER
+prep_build
+build
+make_package
+clean_up
+
+# Vim hints
+# vim:ts=4:sw=4:et:fdm=marker
--- a/build/rlwrap/local.mog
+++ b/build/rlwrap/local.mog
@@ -1,0 +1,20 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+# Copyright 2022 r7st r7st.guru@gmail.com
+
+# Skip dependency detection for python scripts
+<transform file path=.*\.py -> set pkg.depend.bypass-generate .* >
+<transform file path=.*dissect_prompt -> set pkg.depend.bypass-generate .* >
+<transform file path=.*handle_sigwinch -> set pkg.depend.bypass-generate .* >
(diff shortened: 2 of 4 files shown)


oetiker (approved):

build/rlwrap/build.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Copyright 2022 r7st r7st.guru@gmail.com
+
+. ../../lib/build.sh
+
+PROG=rlwrap
+VER=0.45.2
+PKG=ooce/util/rlwrap
+SUMMARY="readline wrapper"
+DESC="rlwrap is a 'readline wrapper', a small utility that uses the GNU Readline library to allow the editing of keyboard input for any command."
```
citrus-it, line 23: Could you please wrap this to 80 columns? A lot of packages do something like this:
```suggestion
DESC="rlwrap is a 'readline wrapper', a small utility that uses the GNU "
DESC+="Readline library to allow the editing of keyboard input for any command."
```

build/rlwrap/build.sh
```diff
@@ -0,0 +1,41 @@
...
+
+set_arch 64
+
+set_mirror "$GITHUB/hanslub42/$PROG/releases/download"
+set_checksum sha256 "9f8870deb46e473d21b5db89d709b6497f4ef9fa06d44eebc5f821daa00c8eca"
```
citrus-it, line 28: I've put the sources on the omnios mirror now, along with a checksum file. Please can you remove these two lines?

build/rlwrap/build.sh
```diff
@@ -0,0 +1,41 @@
...
+
+CFLAGS+=" -fstack-protector -gdwarf"
```
citrus-it, line 30: You shouldn't need to set these explicitly, the framework takes care of adding the appropriate flags for stack protector and DWARF that is compatible with the CTF format, and varies depending on which omnios release is being built for. I just tried this on r151042 and see:

```
/opt/gcc-11/bin/gcc -DHAVE_CONFIG_H -I. -I..    -DDATADIR=\"/opt/ooce/share\" 
    -O2 -fno-aggressive-loop-optimizations
    -fstack-protector-strong -fstack-protector
    -gdwarf -gdwarf-2 -gstrict-dwarf
    -m64 -MT ptytty.o -MD -MP -MF .deps/ptytty.Tpo -c -o ptytty.o ptytty.c
```
so the framework is already adding `-fstack-protector-strong` and `-gdwarf-2 -gstrict-dwarf` and then converting the DWARF to the native CTF data. I think this line should just be deleted.

build/rlwrap/build.sh
```diff
@@ -0,0 +1,41 @@
...
+
+init
+download_source v$VER $PROG-$VER
+prep_build
+build
+strip_install
```
citrus-it, line 36: Please remove this line (`strip-install`). The framework takes care of converting the DWARF to CTF and then removing it from the binary, so it ends up effectively stripped.

build/rlwrap/build.sh
```diff
@@ -0,0 +1,41 @@
...
+
+set_arch 64
+
+set_mirror "$GITHUB/hanslub42/$PROG/releases/download"
+set_checksum sha256 "9f8870deb46e473d21b5db89d709b6497f4ef9fa06d44eebc5f821daa00c8eca"
+
+CFLAGS+=" -fstack-protector -gdwarf"
+
+init
+download_source v$VER $PROG-$VER
```
citrus-it, line 33: With the sources on the omnios mirror, this should change to:
```suggestion
download_source $PROG $PROG $VER
```

build/rlwrap/local.mog
```diff
@@ -0,0 +1,21 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# CDDL HEADER START
```
citrus-it, line 10: This is a stray line
```suggestion
```

citrus-it (changes requested):
Thanks for the PR, and for contributing to OmniOS.

I just have a few changes I'd like to see before it gets merged, please seem the comments in-line.

build/rlwrap/build.sh
r7st: Thanks, I'm able to pull from the mirror.

r7st:
@citrus-it Revised as requested. Please let me know if anything else needs to be modified!

citrus-it (approved):

PR #1171, opened 2022-09-03
Retrieve CTF data from genunix instead of sockfs

citrus-it:
As of `14768: retire nca`, the `sockfs` module no longer contains CTF data for the types that `lsof` needs. However, these are present in `genunix` and apparently always have been (at least back to r151038) so we can safely patch `lsof` to read those. 

```
r151038% ctfdump -c /kernel/amd64/genunix| grep sotpi
typedef struct sotpi_info sotpi_info_t;
struct sotpi_info { /* 0x140 bytes */

r151038% ctfdump -c /kernel/amd64/genunix| grep soaddr
struct soaddr { /* 0x10 bytes */
```

### DIFF
--- a/build/lsof/patches/genunix.patch
+++ b/build/lsof/patches/genunix.patch
@@ -1,0 +1,21 @@
+
+As of "14768: retire nca", the sockfs module no longer contains CTF data for
+the types that lsof needs. However, these are present in genunix and apparently
+always have been (at least back to r151038) so we can safely patch lsof to read
+those instead.
+
+The patch doesn't adjust the 32-bit kernel line since it would be changing
+one non-existent path (/kernel/fs/sockfs) to another.
+
+diff -wpruN '--exclude=*.orig' a~/dialects/sun/dnode.c a/dialects/sun/dnode.c
+--- a~/dialects/sun/dnode.c	1970-01-01 00:00:00
++++ a/dialects/sun/dnode.c	1970-01-01 00:00:00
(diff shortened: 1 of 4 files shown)


build/lsof/patches/genunix.patch
```diff
@@ -0,0 +1,12 @@
...
+--- a~/dialects/sun/dnode.c	1970-01-01 00:00:00
++++ a/dialects/sun/dnode.c	1970-01-01 00:00:00
+@@ -88,7 +88,7 @@ static	int	Sockfs_ctfs = 0;	/* CTF initi
+ 					 * sockfs */
+ 
+ #  if	defined(_LP64)
+-#define	SOCKFS_MOD_FORMAT "/kernel/fs/%s/sockfs"
++#define	SOCKFS_MOD_FORMAT "/kernel/%s/genunix"
+ #  else	/* !defined(_LP64) */
+ #define	SOCKFS_MOD_FORMAT "/kernel/fs/sockfs"
```
sjorge, line 11: This is the fallback for when lsof is !64-bit right?  Should it also grab it from genunix in that case?

oetiker (dismissed):

build/lsof/patches/genunix.patch
citrus-it: Yes, but a 32-bit lsof is for a 32-bit kernel, which we don't have. `/kernel/fs/sockfs` does not exist, and nor does `/kernel/genunix` so I did not bother changing one non-existent path to another.

hadfl (approved):

PR #1185, opened 2022-09-30
Add postgresql 15, drop postgresql 10

mvwieringen:
This can be merged when postgresql 15 gets released which is planned to be at 13/10/2022 I just tested this with the
currect 15rc1 of 29/09/2022. The different extensions already support version 15 (the only one not supporting it was
citus but with the current version that also added 15 to the configure detection. I added a small patch to pg_repack so
it finds some definitions which are in pwd.h on Illumos. Not sure why it wasn't complaining before but with version 15 it
was so I fixed it.


oetiker (dismissed):

mvwieringen:
Rebased and tested against pg15rc2 good to go for release of pg15 next week.

mvwieringen:
15.0 was just released patches are rebased on GA version.

hadfl (approved):

citrus-it (approved):

PR #1213, opened 2022-11-16
First pass at helix pkg

papertigers:
@hadfl / @citrus-it  this is what I pinged about over chat.

Some open issues here:
- I don't think I have `files/hx` setup properly to land in `/opt/ooce/bin/hx`
- The so files in the runtime dir shipped by helix get stripped (not sure if this matters?)
- I have an issue somewhere that's reporting "Manifest contains unresolved variables"

The idea though is that this binary should land in `/opt/ooce/helix/bin/hx` and the runtime lands in `/var/opt/occe/helix/runtime` and the wrapper script exports the proper variable to tell helix where the runtime lives.

This is what linux distros do.  An example of one such case is:
https://github.com/archlinux/svntogit-community/blob/packages/helix/trunk/helix.sh

### DIFF
--- a/build/helix/build.sh
+++ b/build/helix/build.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+
+# Copyright 2022 OmniOS Community Edition (OmniOSce) Association.
+
+. ../../lib/build.sh
+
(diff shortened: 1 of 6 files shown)


papertigers:
Also note that I haven't updated things like `packages.md`

papertigers:
Is there a way to skip the dangling symlink check for specific directories?  The runtime dir that gets built seems to have some such as:
```
-- building package ooce/editor/helix
-- Checking for dangling symlinks
Dangling symlink /var/tmp/build_link/hx-22.12/ooce_editor_helix_pkg/var/opt/ooce/hx/runtime/grammars/sources/twig/node_modules/.bin/tree-sitter
```

papertigers:
For now I pushed a commit that removes the dangling symlinks.  However I think the local.mog file is wrong with the wrapper script because I get:
```
ERROR pkglint.dupaction001.1      path opt/ooce/hx/bin/hx is a duplicate delivered by pkg:/ooce/editor/helix@22.12,5.11-151040.0 under all variant combinations
```

citrus-it (approved):

oetiker (approved):

PR #1332, opened 2023-08-19
Want atuin

papertigers:
This provides [atuin](https://atuin.sh) so that I can stop manually running `cargo install --locked atuin` everywhere.
Once this is in, I plan to update their [install script](https://github.com/atuinsh/atuin/blob/main/install.sh#L105) to use `pkg install atuin`.

I built this locally on r46.
Note that I locally use `MIRROR=https://foo..` in `lib/site.sh` providing the downloaded GitHub release tar and checksum digest on my fileserver.

I am not sure how to gate this on a particular Rust version as building with the rust in r40 was too old prompting me to update my development VM. It seems there is at least some cargo workspace feature being used that is not supported in older Rust releases.

Let me know if there's anything else I can do to help move this PR along.

### DIFF
--- a/build/atuin/build.sh
+++ b/build/atuin/build.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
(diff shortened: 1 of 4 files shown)


hadfl (approved):

citrus-it (approved):

citrus-it:
> I am not sure how to gate this on a particular Rust version as building with the rust in r40 was too old prompting me to update my development VM. It seems there is at least some cargo workspace feature being used that is not supported in older Rust releases.

r40 is out of support which is why the rust package there is old. You can use rustup if you need to compile on that version (or update to a supported release).

PR #1338, opened 2023-09-13
zig: update to 0.11.0

rzezeski:
I'd like to submit this work to update the Zig package to 0.11.0. This upgrade required a bit of work as Zig has transitioned to its self-hosted compiler since the 0.9.1 version. I won't claim that everything works but I see 35,172 of Zig's tests passing when running `zig build test -Dskip-non-native`. My own personal Zig application also compiled and ran. I wanted to try compiling/running ncdu but it requires ncursesw and I don't see that available on our system.

I didn't generate the diffs as described in the instructions. I just used `git diff`. All the patches seem to apply fine. But let me know if you want this changed. After this lands I'll start looking at Zig 0.12.0 dev and start upstreaming patches. I'll also see about getting a proper illumos tag like we have in Go/Rust/etc.


oetiker:
try to make ci build pass

build/zig/build.sh
```diff
@@ -27,7 +27,7 @@ set_arch 64
 # We want to populate the clang-related environment variables
 # and set PATH to point to the correct llvm/clang version
 # but we want to build with gcc for releases before r151041
-set_clangver 13 # zig 0.9.x requires LLVM 13
+set_clangver 16 # zig 0.11.x requires LLVM 16
```
citrus-it, line 30: LLVM 16 is only available in r151045 and higher, so can you add a `min_rel` call so that it just gets skipped on earlier releases?

That would be just `minrel 151045`, before `set_arch` for consistency with other scripts.

@hadfl do you think we should preserve the build script for 0.9.1 for those older releases?

citrus-it (changes requested):
Thank you very much for taking this on and getting this updated. I just have a couple of minor things and could you also please update the version in `doc/packages.md` which should also fix CI.

build/zig/build.sh
rzezeski: Thanks, I also removed the `test_relvar` logic around 151041 since it doesn't seem to apply anymore.

I'm happy to add 0.9.1 back in. Though keep in mind it's an outdated version of the language and stdlib -- Zig has changed a lot in those two point releases.

oetiker (approved):

rzezeski:
I decided this patch needs a bit more work after some discussion with @citrus-it. We'd like to keep some version of Zig available for all supported releases, which means `0.9.1` needs to stick around for a bit longer. Furthermore, `ncdu` is written in Zig and the current version (`2.2.1`) relies on `zig-0.9.1` to build correctly. This is the case because Zig is free to make breaking changes across the board up until the `1.0.0` release. We can't just bump all versions to the latest because Zig tracks latest LLVM, and from my understanding we don't have support for newer LLVMs on older OmniOS releases.

All that said, I'm going to restructure this change so that we have separate `zig9` and `zig11` packages. I think this also makes it easier to support a development package in the form of `zig12`, but I'm not going to do that in this change.

rzezeski:
I see this message during the pkg lint stage when building `zig-0.9` after I've already built `zig-0.11`. It seems like it's referring to the lack of mediation between the two packages.

```
--- Formatting manifest
Do you want to run pkglint? (y/n) y

Lint engine setup...
Starting lint run...

ERROR pkglint.dupaction010.2      path opt/ooce/bin/zig has missing mediator attributes across actions in pkg://local.omnios/ooce/developer/zig-0.11@0.11.0,5.11-151047.0:20230915T041933Z pkg:/ooce/developer/zig-0.9@0.9.1,5.11-151047.0
```

However, their manifests show that there is a mediation in place and I can even install both of them at the same time.

zig-0.9 manifest:
```
link mediator=zig mediator-version=0.9 path=opt/ooce/bin/zig target=../zig-0.9/bin/zig
```

zig-0.11 manifest
```
link mediator=zig mediator-version=0.11 path=opt/ooce/bin/zig target=../zig-0.11/bin/zig
```

```
rpz@kalm:~/src/omnios-extra$ pkg list | grep zig
ooce/developer/zig-0.11 (local.omnios)            0.11.0-151047.0            im-
ooce/developer/zig-0.9 (local.omnios)             0.9.1-151047.0             im-

rpz@kalm:~$ zig version
0.11.0

rpz@kalm:~$ ls -la $(which zig)
lrwxrwxrwx   1 root     root          19 Sep 14 22:47 /opt/ooce/bin/zig -> ../zig-0.11/bin/zig

rpz@kalm:~$ ls /opt/ooce/zig*
/opt/ooce/zig-0.11:
bin  lib

/opt/ooce/zig-0.9:
bin  lib
```

build/zig/build-0.11.sh
```diff
@@ -0,0 +1,130 @@
...
+
+# Copyright 2022 OmniOS Community Edition (OmniOSce) Association.
+
+. ../../lib/build.sh
+
+PROG=zig
+VER=0.11.0
+MAJVER=0
+MINVER=11
+PKG=ooce/developer/zig-0.11
```
hadfl, line 23: We usually don't use dots for versions in package FMRIs. Could you please change this to `zig-011`.

build/zig/build-0.11.sh
```diff
@@ -0,0 +1,130 @@
...
+# }}}
+
+# Copyright 2022 OmniOS Community Edition (OmniOSce) Association.
+
+. ../../lib/build.sh
+
+PROG=zig
+VER=0.11.0
+MAJVER=0
+MINVER=11
```
hadfl, line 22: Can we derive these values from `VER`?

build/zig/build-0.11.sh
```diff
@@ -0,0 +1,130 @@
...
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+
+# Copyright 2022 OmniOS Community Edition (OmniOSce) Association.
```
hadfl, line 15: 2023

build/zig/build-0.9.sh
```diff
@@ -18,7 +18,9 @@
 
 PROG=zig
 VER=0.9.1
-PKG=ooce/developer/zig
+MAJVER=0
+MINVER=9
+PKG=ooce/developer/zig-0.9
```
hadfl, line 23: see above

build/zig/patches-0.11/hard-code-dynamic-linker.patch
```diff
@@ -0,0 +1,47 @@
...
++                .solaris,
+                 => return .none,
+             }
+         }
+@@ -1572,7 +1572,7 @@ pub const Target = struct {
+             .netbsd => return copy(&result, "/libexec/ld.elf_so"),
+             .openbsd => return copy(&result, "/usr/libexec/ld.so"),
+             .dragonfly => return copy(&result, "/libexec/ld-elf.so.2"),
+-            .solaris => return copy(&result, "/lib/64/ld.so.1"),
++            .solaris => return copy(&result, "/usr/lib/amd64/ld.so.1"),
```
hadfl, line 26: What's the reason for this change?

hadfl (changes requested):
Thank you very much for your work on `zig`. This is great!

There are just a few nits which I addressed in the review. Can you on top of that please do the following two things:
- add the package FMRIs to `doc/baseline`
- create a "rename package" for `zig` to point to `zig-09` (you can find an example here: https://github.com/omniosorg/omnios-extra/blob/master/build/obsolete/developer-llvm-130.p5t )

build/zig/build-0.11.sh
rzezeski: Done.

build/zig/build-0.11.sh
rzezeski: Done. You might want to check the clang script. I originally based by design off of that, but it doesn't actually extract the `MINVER` correctly.

build/zig/patches-0.11/hard-code-dynamic-linker.patch
rzezeski: This is the path other executables use for the dynamic linker. So it seemed to me it's the preferred path to use to allow the system flexibility to link it to the correct interpreter.

build/zig/build-0.9.sh
rzezeski: Done.

rzezeski:
@hadfl I've made the changes you requested. I still need to build/test on my end with these changes but I wanted to push it out for review first. You'll definitely want to double-check my work around the obsolete packaging stuff and make sure I didn't mess that up.

I also added another patch to make sure the zig compiler uses frame pointers even when built in an optimized release mode.

rzezeski:
Both packages build with the latest changes. I can install both versions at the same time.

```
rpz@kalm:~/src/omnios-extra$ pkg list | grep zig
ooce/developer/zig-011 (local.omnios)             0.11.0-151047.0            im-
ooce/developer/zig-09 (local.omnios)              0.9.1-151047.0             im-

rpz@kalm:~/src/omnios-extra$ zig version
0.11.0

rpz@kalm:~/src/omnios-extra$ ls -la $(which zig)
lrwxrwxrwx   1 root     root          19 Sep 18 17:13 /opt/ooce/bin/zig -> ../zig-0.11/bin/zig
```

I still see a failure during `pkglint` and I have no idea why.

```
--- Formatting manifest
Do you want to run pkglint? (y/n) y

Lint engine setup...
Starting lint run...

ERROR pkglint.dupaction010.2      path opt/ooce/bin/zig has missing mediator attributes across actions in pkg://local.omnios/ooce/developer/zig-011@0.11.0,5.11-151047.0:20230918T191625Z pkg:/ooce/developer/zig-09@0.9.1,5.11-151047.0
----- pkglint failed
```

I'm not sure how to test the package renaming.

citrus-it (commented):
The mediator warning is something you can ignore. When it's doing the `pkglint`, it's checking the new package against the existing published ones, and the old `zig` package delivers `opt/ooce/bin/zig` as a link, but that link doesn't have any mediator attributes on it, hence the conflict.

It isn't taking the renamed package into account because it's not published yet. This should be fine once everything is published.

build/meta/extra-build-tools.p5m
```diff
@@ -25,7 +25,7 @@ depend fmri=ooce/developer/llvm-14 type=require
 depend fmri=ooce/developer/clang-13 type=require
 depend fmri=ooce/developer/subversion type=require
 depend fmri=ooce/developer/yasm type=require
-depend fmri=ooce/developer/zig type=require
+depend fmri=ooce/developer/zig-0.9 type=require
```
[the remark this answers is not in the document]
rzezeski: Done.

build/ncdu/build.sh
```diff
@@ -27,7 +27,7 @@ DESC="Disk usage analyzer with an ncurses interface"
 set_arch 64
 set_clangver # zig requires CC to be set
 
-BUILD_DEPENDS_IPS="ooce/developer/zig"
+BUILD_DEPENDS_IPS="ooce/developer/zig-0.9"
```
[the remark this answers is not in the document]
rzezeski: Done.

build/obsolete/build.sh
```diff
@@ -28,6 +28,8 @@ PKG=ooce/obsolete/developer/clang-130
 test_relver '>=' 151036 && publish_manifest "" developer-clang-130.p5t
 PKG=ooce/obsolete/developer/llvm-130
 test_relver '>=' 151036 && publish_manifest "" developer-llvm-130.p5t
+PKG=ooce/obsolete/developer/zig
```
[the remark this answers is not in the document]
rzezeski: Done.

build/obsolete/developer-zig.p5t
```diff
@@ -0,0 +1,3 @@
+set name=pkg.fmri value=pkg://$(PKGPUBLISHER)/ooce/developer/zig@0.9.1,$(SUNOSVER)-$(PVER)
```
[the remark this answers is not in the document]
rzezeski: Done.

rzezeski:
> The mediator warning is something you can ignore. When it's doing the `pkglint`, it's checking the new package against the existing published ones, and the old `zig` package delivers `opt/ooce/bin/zig` as a link, but that link doesn't have any mediator attributes on it, hence the conflict.
> 
> It isn't taking the renamed package into account because it's not published yet. This should be fine once everything is published.

You're right, taking a closer look at the error you can see the `zig-011` FMRI comes from `local.omnios` whereas the `zig-09` comes from the existing `ooce` repo. It fooled me because I would have expected it to still use the old package name `zig`, but I guess it ends up using the new name instead.

Anyways, I think I fixed the other issues you mentioned.

build/zig/build-0.11.sh
hadfl: The variable name `MINVER` in the clang build script is a bit misleading as what we want there is `$MAJVER.$MINVER`. We should probably consider changing `MINVER` to `MAJMINVER` to avoid confusion. Sorry for this.

hadfl (approved):

hadfl (approved):

The-King-of-Toasters:
Thank you @rzezeski! I'm the author of the initial Illumos port, but I've been struggling to keep it working on my own. I'll try to prepare a pr to get these patches upstream. Note that Zig has recently changed over to llvm 17, so there may be build failures if you start packaging 0.12.

rzezeski:
@The-King-of-Toasters If you don't mind I'd like to get these upstreamed myself so I can engage the Zig community with illumos support going forward. I want to work towards getting a proper illumos tag and making sure various bits work.

The-King-of-Toasters:
Of course!

My dream when starting the port was to have Illumos be at least a Tier 2 supported system like FreeBSD. This requires a CI system and proper support for evented IO using evports that I never got around to add.

rzezeski:
@The-King-of-Toasters Thanks. And thank you for your initial work, otherwise I might not have done this work at all.

I'm in the process of bringing up 0.12.0-dev as we speak and that's what I'm going to base my initial upstream patches on.

PR #1355, opened 2023-10-12
fix regression in unbound 1.18

hadfl:


sjorge:
I tested this and unbound now works again to resolve over ipv4 and ipv6.

citrus-it (approved):

PR #1394, opened 2024-01-18
Add gptfdisk 1.0.9

luqmana:
Packages up the [GPT fdisk](https://sourceforge.net/projects/gptfdisk/) suite of tools: `gdisk`, `cgdisk`, `sgdisk`, `fixparts`.

### DIFF
--- a/build/gptfdisk/build.sh
+++ b/build/gptfdisk/build.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+
+# Copyright 2021 OmniOS Community Edition (OmniOSce) Association.
+# Copyright 2024 Oxide Computer Company
+
+. ../../lib/build.sh
+
+PROG=gptfdisk
+VER=1.0.9
+PKG=ooce/system/gptfdisk
+SUMMARY="GPT fdisk (gdisk, cgdisk, sgdisk, fixparts)"
+DESC="GPT fdisk is a disk partitioning tool loosely modeled on Linux "
+DESC+="fdisk, but used for modifying GUID Partition Table (GPT) disks."
+
+POPTVER=1.14
+
+set_arch 64
+
+init
+prep_build
+
+#########################################################################
+# Download and build a static version of popt
+
+save_buildenv
+
+CONFIGURE_OPTS=" --disable-shared --enable-static"
+
+build_dependency popt popt-$POPTVER popt popt $POPTVER
+
+restore_buildenv
+
+export POPT_CFLAGS="-I$DEPROOT/$PREFIX/include"
+export POPT_LDFLAGS="-L$DEPROOT/$PREFIX/lib/amd64"
+
+#########################################################################
+
+pre_configure() {
+    typeset arch=$1
+
+    MAKE_ARGS_WS="
+        CXXFLAGS=\"$CXXFLAGS ${CXXFLAGS[$arch]} $POPT_CFLAGS\"
+        LDFLAGS=\"$LDFLAGS ${LDFLAGS[$arch]} $POPT_LDFLAGS -zignore\"
+    "
+
+    # No configure
+    false
+}
+
+make_install() {
+    for d in bin share/man/man8; do
+        logcmd $MKDIR -p $DESTDIR$PREFIX/$d
+    done
+
+    for p in gdisk cgdisk sgdisk fixparts; do
+        logcmd $CP $TMPDIR/$BUILDDIR/$p \
+            $DESTDIR$PREFIX/bin/$p
+        logcmd $CP $TMPDIR/$BUILDDIR/$p.8 \
+            $DESTDIR$PREFIX/share/man/man8/$p.8
+    done
+}
+
+download_source $PROG $PROG $VER
+patch_source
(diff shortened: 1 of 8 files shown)


build/gptfdisk/build.sh
```diff
@@ -0,0 +1,84 @@
...
+CONFIGURE_OPTS=" --disable-shared --enable-static"
+
+build_dependency popt popt-$POPTVER popt popt $POPTVER
+
+restore_buildenv
+
+export POPT_CFLAGS="-I$DEPROOT/$PREFIX/include"
+export POPT_LDFLAGS="-L$DEPROOT/$PREFIX/lib/amd64"
+
+#########################################################################
```
citrus-it, line 47: Since you copied this across from `exif/build.sh`, can you please also include the copyright line from that file at the top of this one (above the new Oxide copyright)?

build/gptfdisk/build.sh
```diff
@@ -0,0 +1,84 @@
...
+pre_configure() {
+    typeset arch=$1
+
+    MAKE_ARGS_WS="
+        CXXFLAGS=\"$CXXFLAGS ${CXXFLAGS[$arch]} $POPT_CFLAGS\"
+        LDFLAGS=\"$LDFLAGS ${LDFLAGS[$arch]} $POPT_LDFLAGS -zignore\"
+    "
+
+    # No configure
+    false;
```
citrus-it, line 61: You can drop the trailing semi-colon here.

build/gptfdisk/build.sh
```diff
@@ -0,0 +1,84 @@
...
+
+restore_buildenv
+
+export POPT_CFLAGS="-I$DEPROOT/$PREFIX/include"
+export POPT_LDFLAGS="-L$DEPROOT/$PREFIX/lib/amd64"
+
+#########################################################################
+
+set_mirror "https://download.sourceforge.net/project/gptfdisk"
+set_checksum sha256 "dafead2693faeb8e8b97832b23407f6ed5b3219bc1784f482dd855774e2d50c2"
```
citrus-it, line 50: This is useful during package development but I've put a copy of this source on the omnios download mirror; please can you remove these two lines now?

There are two main reasons for this:
- most software licences require that we make the source available for any binaries that we ship so to make this simpler we maintain a local archive where possible;
- Having all of the sources in one place allows for easier caching and building software offline (we don't always completely achieve this with some languages that love to pull things from the Internet, but we try).

build/gptfdisk/build.sh
```diff
@@ -0,0 +1,84 @@
...
+
+    for p in gdisk cgdisk sgdisk fixparts; do
+        logcmd $CP $TMPDIR/$BUILDDIR/$p \
+            $DESTDIR$PREFIX/bin/$p
+        logcmd $CP $TMPDIR/$BUILDDIR/$p.8 \
+            $DESTDIR$PREFIX/share/man/man8/$p.8
+    done
+}
+
+download_source $PROG/$VER $PROG $VER
```
citrus-it, line 77: With the removal of the set_mirror/checksum lines above:
```suggestion
download_source $PROG $PROG $VER
```

doc/packages.md
```diff
@@ -195,6 +195,7 @@
 | ooce/storage/minio		| 2024-01-13T07-53-03Z | https://github.com/minio/minio/releases | [omniosorg](https://github.com/omniosorg)
 | ooce/storage/minio-mc		| 2024-01-13T08-44-48Z | https://github.com/minio/mc/releases | [omniosorg](https://github.com/omniosorg)
 | ooce/system/file-system/ntfs-3g | 2022.10.3	| https://github.com/tuxera/ntfs-3g/releases | [omniosorg](https://github.com/omniosorg)
+| ooce/system/gptfdisk		| 1.0.9		| https://sourceforge.net/projects/gptfdisk/files/gptfdisk/ | [omniosorg](https://github.com/omniosorg)
```
citrus-it, line 198: The last field here is the maintainer, which could be Oxide if that's the intention. No requirement, it can be left as omniosorg too.

build/gptfdisk/patches/no-icu-libs.patch
```diff
@@ -0,0 +1,14 @@
...
+--- a~/Makefile	1970-01-01 00:00:00
++++ a/Makefile	1970-01-01 00:00:00
+@@ -131,8 +131,8 @@ else ifeq ($(TARGET),macos)
+   LDLIBS= #-licucore
+   CGDISK_LDLIBS=/usr/local/Cellar/ncurses/6.2/lib/libncurses.dylib
+ else ifeq ($(TARGET),solaris)
+-  CXXFLAGS+=-Wall -D_FILE_OFFSET_BITS=64 -I/usr/include/ncurses
+-  LDFLAGS+=-L/lib -licuio -licuuc -luuid
++  #CXXFLAGS+=-Wall -D_FILE_OFFSET_BITS=64 -I/usr/include/ncurses
++  #LDFLAGS+=-L/lib -licuio -licuuc -luuid
```
citrus-it, line 11: I wonder why it's save to remove `-luuid` here. It builds and runs so it must come in from somewhere else.

citrus-it (changes requested):
Thanks for the contribution, this looks really good. I just have a few comments and suggestions.

build/gptfdisk/build.sh
luqmana: added.

build/gptfdisk/build.sh
luqmana: thanks! swapped over to omnios mirror

build/gptfdisk/patches/no-icu-libs.patch
luqmana: Yup, `LDFLAGS` is seeded with `-luuid` just [above](https://sourceforge.net/p/gptfdisk/code/ci/master/tree/Makefile#l71).

hadfl (dismissed):

citrus-it (approved):

citrus-it (approved):

PR #1411, opened 2024-02-19
caddy: add version 2.7.6 to extra.omnios

gkoh:
Add v2.7.6 of the Caddy web server.

Add patch to fix Solaris-derivative build.
Comes from caddy master:76611fa15079b755ddf3c859ae2ea0ad2123223f and this PR 6021: https://github.com/caddyserver/caddy/pull/6021

### DIFF
--- a/build/caddy/build.sh
+++ b/build/caddy/build.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+# }}}
+
+# Copyright 2024 Guo-Rong Koh
+
+. ../../lib/build.sh
+
+PROG=caddy
+VER=2.7.6
+PKG=ooce/server/caddy
+SUMMARY="Caddy web server"
+DESC="Fast and extensible multi-platform HTTP/1-2-3 web server with automatic \
+HTTPS"
+
+CONFIG=etc/${PREFIX#/}/$PROG
+DATA=var/${PREFIX#/}/$PROG
+
+RUN_DEPENDS_IPS="ooce/server/webservd-common"
+
+set_arch amd64
+set_gover
+
+XFORM_ARGS="
+    -DPREFIX=${PREFIX#/}
+    -DPROG=$PROG
+    -DVERSION=$VER
+    -DUSER=webservd -DGROUP=webservd
+    -DCONFIG=$CONFIG
+    -DDATA=$DATA
+    -DXDG_CONFIG=${CONFIG%/$PROG}
+    -DXDG_DATA=${DATA%/$PROG}
+"
+
+build() {
+    pushd $TMPDIR/$BUILDDIR > /dev/null
+
+    logmsg "Building $PROG"
+    export CGO_ENABLED=0
+    export GOOS=illumos
+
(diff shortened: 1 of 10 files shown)


gkoh:
The SmartOS pkgsrc repository dropped caddy for some reason, so I've opted to add it to the OmniOS extra IPS instead.

hadfl (commented):
Thank you very much for your contribution. This looks good so far. However, since this is a service we should provide an SMF service manifest. Have a look at other stuff in omnios-extra that provides a service manifest.

Let me know if you get stuck or have questions and we can help getting this done and integrated.

gkoh:
> Thank you very much for your contribution. This looks good so far. However, since this is a service we should provide an SMF service manifest. Have a look at other stuff in omnios-extra that provides a service manifest.
> 
> Let me know if you get stuck or have questions and we can help getting this done and integrated.

Done! Thank you for the initial review.
I selected a random available uid/gid from `doc/idlist.md`, this could well reuse `webservd` though.

build/caddy/build.sh
```diff
@@ -0,0 +1,70 @@
...
+DATA=var/${PREFIX#/}/$PROG
+
+set_arch amd64
+set_gover
+
+XFORM_ARGS="
+    -DPREFIX=${PREFIX#/}
+    -DPROG=$PROG
+    -DVERSION=$VER
+    -DsVERSION=
```
hadfl, line 36: Given that you set this empty, i.e. the service is not versioned, you might as well drop this parameter and remove it in the manifest and local.mog

hadfl (commented):
> Done! Thank you for the initial review. I selected a random available uid/gid from `doc/idlist.md`, this could well reuse `webservd` though.

Yeah, I think re-using `webservd` would be a good idea (like we do for apache). Given we are limited to IDs below 100 for service accounts we should not allocate new ones if not necessary.

build/caddy/build.sh
gkoh: Done.

gkoh:
> > Done! Thank you for the initial review. I selected a random available uid/gid from `doc/idlist.md`, this could well reuse `webservd` though.
> 
> Yeah, I think re-using `webservd` would be a good idea (like we do for apache). Given we are limited to IDs below 100 for service accounts we should not allocate new ones if not necessary.

I have made this change locally, but not sure how to proceed.
If I add the webservd uid/gid in the caddy pkg, it results in conflicts with the apache pkg.
If I don't add the webservd uid/gid, then it requires apache to be installed just to have the uid/gid present.

Can you recommend a way forward? (I couldn't find a way to optionally create a user).

hadfl:
> Can you recommend a way forward? (I couldn't find a way to optionally create a user).

So the way forward is to create a meta package that will ship the user/group, then make apache and caddy depend on it. We do this already for `nginx` as we ship multiple versions there. We'd have to do this anyway for `webservd` once we ship another apache version.

I quickly prepared a commit (untested) that should show you the principle. Feel free to copy whatever you need from it:
https://github.com/hadfl/omnios-extra/commit/544e1ba5d44d9365b88cbe049424843d3cea37fd

gkoh:
> > Can you recommend a way forward? (I couldn't find a way to optionally create a user).
> 
> So the way forward is to create a meta package that will ship the user/group, then make apache and caddy depend on it. We do this already for `nginx` as we ship multiple versions there. We'd have to do this anyway for `webservd` once we ship another apache version.
> 
> I quickly prepared a commit (untested) that should show you the principle. Feel free to copy whatever you need from it: [hadfl@544e1ba](https://github.com/hadfl/omnios-extra/commit/544e1ba5d44d9365b88cbe049424843d3cea37fd)

Aha, thank you for that.
I have updated, but it is untested.

How does one build this meta package for testing?

gkoh:
> > How does one build this meta package for testing?
> 
> You can use `buildctl` in the `build` directory to build any package (including meta packages). In this case: `./buildctl build webservd-common`

Perfect, thanks.
Done and tested, confirmed working as expected.

hadfl (approved):

PR #1425, opened 2024-03-18
cyrus-imapd: update to 3.8.2

mvwieringen:
Upstream have reverted ssl patch so we can drop our local patch.

It also seems man pages are installed now and as such the package diff shows local man pages and symlinks to /opt/ooce/man/man?/ which I think is fine.


citrus-it:
Thanks for taking care of this.

> It also seems man pages are installed now and as such the package diff shows local man pages and symlinks to /opt/ooce/man/man?/ which I think is fine.

Yes, me too. Nice improvement.

citrus-it (approved):

oetiker (approved):

PR #1451, opened 2024-04-29
zig: update to 0.12.0

rzezeski:


hadfl (commented):
Thank you so much for taking care of keeping `zig` up to date on OmniOS. This looks good, just a few nits:

- Since this is a new package, could you add it to `doc/packages.md` and `doc/baseline`
- This requires clang 17 as a build time dependency for releases 151045 and later. Could you please adjust the comment and minimum release in the `extra-build-tools` meta package: https://github.com/omniosorg/omnios-extra/blob/master/build/meta/extra-build-tools.p5m#L128-L130

rzezeski:
Whoops, thanks for catching those. Do you prefer updates to be additional commits or would you rather a force push?

hadfl:
> Whoops, thanks for catching those. Do you prefer updates to be additional commits or would you rather a force push?

I think for this one you might just force push it. Thanks!

rzezeski:
Okay, hopefully I got that right.

rzezeski:
ping

hadfl (approved):

citrus-it (approved):

hadfl:
> Okay, hopefully I got that right.

I did a small change to `extra-build-tools`. My comment was maybe a bit misleading. We don't require zig 0.12, yet to build anything so it does not need to be in the meta package. However, we require clang 17 to build zig 0.12 and since we build zig from r151047 on we need to change the condition for clang 17 to that version.

PR #1469, opened 2024-06-05
dont-package-helix-grammar-sources

snltd:
Omit the grammar sources from the Helix package. This removes the better part of 20,000 small files from the package, none of which are not required to run Helix. 

[Homebrew already do this](https://github.com/helix-editor/helix/issues/3437) and it is [recommended by the Helix devs](https://github.com/helix-editor/helix/issues/6187#issuecomment-1769860105).

### DIFF
--- a/build/helix/local.mog
+++ b/build/helix/local.mog
@@ -17,6 +17,7 @@
 <transform file elfhash path=$(PREFIX)/bin/hx -> edit path bin/hx bin/hx.bin>
 file files/hx path=X/$(PREFIX)/bin/hx mode=0555
 <transform file path=X/ -> edit path X/ "">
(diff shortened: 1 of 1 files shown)


oetiker (approved):

hadfl (approved):

rahji:
The instructions for [building from source](https://docs.helix-editor.com/building-from-source.html) should mention this. I copied the `runtime` folder to `~/.config/helix/` instead of linking it, so I ended up with *2.8GB* of files that I do not need to run helix. It's not obvious to all that the `runtime/sources` folder can be removed. Considering the process for building from source has been made so easy, it would be worthwhile to consider that someone who has never built anything from source could be the audience of those instructions.. and would benefit from knowing that the `sources` folder can be removed after the build process is done.

PR #1492, opened 2024-07-25
node-22: fix build on r151046

hadfl:


build/nodejs/build-22.sh
```diff
@@ -26,6 +26,9 @@ DESC+="servers."
 
 MAJVER=${VER%%.*}
 
+# fails to build with gcc 12
+((GCCVER = 12)) && set_gccver 11
```
oetiker, line 30: do you actually want to assign ? GCCVER to 12 ? If not, use two ==

```suggestion
((GCCVER == 12)) && set_gccver 11
```

hadfl: That was definitely an oversight, thanks for catching it.

citrus-it (approved):

oetiker (approved):

PR #1512, opened 2024-08-21
Add VictoriaMetrics package, a time series database.

gkoh:
Package multiple pieces of the Prometheus compatible VictoriaMetrics suite:
- victoria-metrics
- vmagent
- vmbackup
- vmrestore
- vmctl

Whilst here, correct a minor spelling error in functions.sh.

### DIFF
--- a/build/meta/extra-build-tools.p5m
+++ b/build/meta/extra-build-tools.p5m
@@ -96,10 +96,6 @@ depend fmri=ooce/x11/library/libx11 type=require
 depend fmri=ooce/x11/library/libxau type=require
 depend fmri=ooce/x11/library/libxcb type=require
 
-# clang 14 is the default clang compiler from r151043 onwards
-depend fmri=ooce/developer/clang-14 type=conditional \
-    predicate=release/name@0.5.11-151043
-
 # Added llvm/clang 15 from r151043 onwards
 depend fmri=ooce/developer/llvm-15 type=conditional \
     predicate=release/name@0.5.11-151043


gkoh:
I need some advice on this package, there are some oddities.

1. I have claimed uid/gid 93, hope this is correct

2. Configuration is not via file, it is via command-line argument or equivalent envvar.
Updating the service manifest to change the command line is crazy, so I think the right answer is by applying service profiles that modify the instance method_environment.
I have placed examples of such profiles in `share/victoriametrics/xxx-profile.xml`.

3. Using an envvar, I have set the storageDataPath for the server `victoria-metrics` in the default svc manifest, however I believe this will be overwritten by an update. This should probably be applied via profile only. By doing so, the server and agent will _require_ a service profile to be applied or they will not function correctly.

4. I have added some install notes to cover the profile stuff above, I hope this is sufficient.

build/victoriametrics/build.sh
```diff
@@ -0,0 +1,72 @@
...
+
+PROG=victoriametrics
+VER=1.102.1
+PKG=ooce/database/victoriametrics
+SUMMARY="VictoriaMetrics"
+DESC="Fast, cost-effective monitoring solution and time series database."
+
+DATA=var/lib/victoria-metrics-data
+
+set_arch amd64
```
citrus-it, line 27: While there's no requirement to get this building on omnios aarch64, we do try and make all packages fairly agnostic. Please just specify `64` bits here.
```suggestion
set_arch 64
```

build/victoriametrics/build.sh
```diff
@@ -0,0 +1,72 @@
...
+
+. ../../lib/build.sh
+
+PROG=victoriametrics
+VER=1.102.1
+PKG=ooce/database/victoriametrics
+SUMMARY="VictoriaMetrics"
+DESC="Fast, cost-effective monitoring solution and time series database."
+
+DATA=var/lib/victoria-metrics-data
```
citrus-it, line 25: This should really be `var/$PREFIX/victoriametrics/` - see [filesystem(7)](https://man.omnios.org/filesystem)

build/victoriametrics/build.sh
```diff
@@ -0,0 +1,72 @@
...
+"
+
+build() {
+    pushd $TMPDIR/$BUILDDIR > /dev/null
+
+    logmsg "Building $PROG"
+    export CGO_ENABLED=0
+    export GOOS=illumos
+
+    logcmd gmake victoria-metrics-pure || logerr "Unable to build victoria-metrics-pure"
```
citrus-it, line 48: Please can you wrap this line to bring it under 80 characters, and use `$MAKE`?
```suggestion
    logcmd $MAKE victoria-metrics-pure \
        || logerr "Unable to build victoria-metrics-pure"
```

build/victoriametrics/build.sh
```diff
@@ -0,0 +1,72 @@
...
+prep_build
+build
+install_go bin/victoria-metrics-pure victoria-metrics
+install_go bin/vmagent-pure vmagent
+install_go bin/vmbackup-pure vmbackup
+install_go bin/vmctl-pure vmctl
+install_go bin/vmrestore-pure vmrestore
+xform files/$PROG-template.xml > $TMPDIR/$PROG.xml
+install_smf application $PROG.xml
+xform files/victoria-metrics-profile-template.xml > $TMPDIR/victoria-metrics-profile.xml
```
citrus-it, line 66:
```suggestion
xform files/victoria-metrics-profile-template.xml \
    > $TMPDIR/victoria-metrics-profile.xml
```

build/victoriametrics/files/victoria-metrics-profile-template.xml
```diff
@@ -0,0 +1,40 @@
...
+    source. A copy of the CDDL is also available via the Internet at
+    http://www.illumos.org/license/CDDL.
+
+    Copyright 2024 OmniOS Community Edition (OmniOSce) Association.
+
+-->
+<service_bundle type="profile"
+                name="victoriametrics">
+
+    <service name="application/victoriametrics"
```
citrus-it, line 20:
```suggestion
    <service name="ooce/application/victoriametrics"
```

build/victoriametrics/files/victoriametrics-template.xml
```diff
@@ -0,0 +1,162 @@
...
+    source. A copy of the CDDL is also available via the Internet at
+    http://www.illumos.org/license/CDDL.
+
+    Copyright 2024 OmniOS Community Edition (OmniOSce) Association.
+
+-->
+<service_bundle type="manifest"
+                name="victoriametrics">
+
+    <service name="application/victoriametrics"
```
citrus-it, line 20:
```suggestion
    <service name="ooce/application/victoriametrics"
```

build/victoriametrics/files/victoriametrics-template.xml
```diff
@@ -0,0 +1,162 @@
...
+            <dependency name="filesystem_local"
+                        grouping="require_all"
+                        restart_on="none"
+                        type="service">
+                <service_fmri value="svc:/system/filesystem/local:default" />
+            </dependency>
+            <method_context>
+                <method_credential user="$(USER)"
+                                   group="$(GROUP)"
+                                   privileges="basic,net_privaddr" />
```
citrus-it, line 119: Does this service need to bind to TCP/UDP ports below 1024? If not, please drop the `net_privaddr` privilege.
Same for the other services.

lib/functions.sh
```diff
@@ -219,7 +219,7 @@ logerr() {
     # Print an error message and ask the user if they wish to continue
     logmsg -e "$@" >> /dev/stderr
     if [ -z "$BATCH" ]; then
-        ask_to_continue "An Error occured in the build. "
+        ask_to_continue "An Error occurred in the build. "
```
citrus-it, line 222: Thanks for fixing this.

citrus-it (changes requested):
> I need some advice on this package, there are some oddities.
> 
> 1. I have claimed uid/gid 93, hope this is correct

Yes, and thank you for updating the idlist file. That's where we track allocated UIDs/GIDs across all repositories.
Please could you use tabs instead of spaces so that the new lines match the existing ones?
Same in `packages.md`.

> 2. Configuration is not via file, it is via command-line argument or equivalent envvar.
>    Updating the service manifest to change the command line is crazy, so I think the right answer is by applying service profiles that modify the instance method_environment.
>    I have placed examples of such profiles in `share/victoriametrics/xxx-profile.xml`.

That seems like a good solution, and thank for you for the installation notes too.

> 3. Using an envvar, I have set the storageDataPath for the server `victoria-metrics` in the default svc manifest, however I believe this will be overwritten by an update. This should probably be applied via profile only. By doing so, the server and agent will _require_ a service profile to be applied or they will not function correctly.

Have you checked if they are overwritten on update? I don't think they will be if they have ever been modified with `svccfg` or a profile.

> 4. I have added some install notes to cover the profile stuff above, I hope this is sufficient.

Yes, thank you!

Please could you also update `doc/baseline.md`?

build/victoriametrics/files/victoriametrics-template.xml
gkoh: Good question, they do not. By default both services bind to 8428 and 8429.
Privilege dropped.

gkoh:
> > I need some advice on this package, there are some oddities.
> > 
> > 1. I have claimed uid/gid 93, hope this is correct
> 
> Yes, and thank you for updating the idlist file. That's where we track allocated UIDs/GIDs across all repositories. Please could you use tabs instead of spaces so that the new lines match the existing ones? Same in `packages.md`.

Done.
  
> > 3. Using an envvar, I have set the storageDataPath for the server `victoria-metrics` in the default svc manifest, however I believe this will be overwritten by an update. This should probably be applied via profile only. By doing so, the server and agent will _require_ a service profile to be applied or they will not function correctly.
> 
> Have you checked if they are overwritten on update? I don't think they will be if they have ever been modified with `svccfg` or a profile.

OK, I checked again, properly. You are correct.
On update, values modified by a profile are not overwritten.
Thus, I have left it as-is.
 
> > 4. I have added some install notes to cover the profile stuff above, I hope this is sufficient.
> 
> Yes, thank you!
> 
> Please could you also update `doc/baseline.md`?

I updated both `doc/baseline` and `doc/baseline.aarch64`.
I can't test an aarch64 build (no hardware) but VictoriaMetrics themselves release arch64 binaries.

Thank you for the review.

gkoh:
Apologies, github didn't like my fat-fingered rebase so some of the review comment context got nuked.
I think I have addressed all review comments.

hadfl:
I am very sorry, I messed up when rebasing this which closed the PR. I have re-opened it https://github.com/omniosorg/omnios-extra/pull/1531

Issue #1539, opened 2024-11-05
neovim crashes with "E41: Out of memory!"


jclulow:
@pfmooney and I have seen random crashes from Neovim 0.10 lately, and I have finally had some cycles to dig in and debug it.  The crux of the issue is that the embedded LuaJIT library is doing some relatively daft things with mappings, and depending on how unlucky you get with random numbers, it will collide with the process heap and **malloc()** will stop being able to allocate more memory.

When this happens, the process exits unceremoniously with somewhat corrupted output:

```
E41: Out of memory!
                   Vim: Finished.
$
```

I used DTrace to get a sense of what was causing the allocation failures.  They're all failures from **malloc()**, and if you look at a regular **nvim** process that has not yet failed, it's not hard to see why:

```
$ pmap -x $(pgrep -f /bin/nvim.--embed)
27112:  /opt/ooce/neovim/bin/nvim --embed
         Address     Kbytes        RSS       Anon     Locked Mode   Mapped File
0000000000400000       5396       4560          -          - r-x--  nvim
0000000000954000         92         92         64          - rw---  nvim
000000000096B000        104         92         92          - rw---  nvim
0000000001204000       7152       7144       7144          - rw---    [ heap ]
000000000B3C0000         64         56          -          - r-x--    [ anon ]
000000000B3D0000         64         64          -          - r-x--    [ anon ]
00007396754A3000       2048       1964       1964          - rw---    [ anon ]
FFFFFC7FECA12000        324        288          -          - r-x--  ld.so.1
FFFFFC7FECA73000         12         12         12          - rwx--  ld.so.1
FFFFFC7FECA76000          8          8          8          - rwx--  ld.so.1
FFFFFC7FECE90000          4          4          4          - rwx--    [ anon ]
...
```

Critically, in this particular instance, the heap currently extends up to `0x1205BF0`, and the `MAP_ANON` mappings that LuaJIT has created are as low as `0xB3C0000`, which is only 161MB away!  The initial address that's chosen for the mapping is made up by a random number generator in LuaJIT, and constrained to be within about 2GB of some of the language runtime functions; these are compiled into the binary, so they're down _under_ the heap in the address space.  Further compounding the issue, LuaJIT appears to really want to grow its allocation _downward_, setting it on a collision course with the heap.

[...]

PR #1573, opened 2025-01-22
neovim: update to 0.10.3, work around #1539

jclulow:
This change updates Neovim to 0.10.3, which includes a relatively crucial LSP bug fix around server cancellation.  It also adds a temporary workaround for #1539, until we can find time to fix the real issues there.

I have built and tested this on an OmniOS r151046 LTS system, and it appears to work.  I ran ":checkhealth" and also the plugin checks for the Lazy plugin system, and there were no weird new problems.  I also edited a text file and made use of the RPC socket.

### DIFF
--- a/build/neovim/build.sh
+++ b/build/neovim/build.sh
@@ -17,7 +17,7 @@
 . ../../lib/build.sh
 
 PROG=neovim
-VER=0.10.2
+VER=0.10.3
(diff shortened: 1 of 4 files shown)


citrus-it (approved):
Thanks!

oetiker (approved):

PR #1586, opened 2025-02-21
postgresql package updates

mvwieringen:
CVE-2025-1094  and drop unsupported version 12.


oetiker (approved):

hadfl (changes requested):
Since you are removing postgresql 12, you'll also have to remove it from doc/baseline and the extra-build-tools meta package.

hadfl (commented):
There is also a leftover `ooce/library/postgresql-12` in doc/baseline

hadfl (approved):

PR #1592, opened 2025-03-20
add zig-0.14.0

rzezeski:
There's some issue with threadlocal storage, but I don't think it needs to hold up the package given zig's quickly evolving nature. I might look into it more in the coming weeks and perhaps a follow-up patch could be done.

### DIFF
--- a/build/zig/build-0.14.sh
+++ b/build/zig/build-0.14.sh
@@ -1,0 +1,122 @@
+#!/usr/bin/bash
+#
+# {{{ CDDL HEADER
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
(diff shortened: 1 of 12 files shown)


build/zig/build-0.14.sh
```diff
@@ -0,0 +1,122 @@
...
+VER=0.14.0
+MAJVER=${VER%%.*}
+MINVER=${VER%.*}
+MINVER=${MINVER#*.}
+PKG=ooce/developer/zig-014
+SUMMARY="$PROG programming language"
+DESC="$PROG is a general-purpose programming language and toolchain for "
+DESC+="maintaining robust, optimal, and reusable software."
+
+min_rel 151049
```
hadfl, line 29: we only ship llvm/clang 19 from r151051 on. this build will likely break on r150150 which is still supported. the barrier should probably be r52

hadfl (commented):
This looks good. Thanks for your ongoing work for zig on OmniOS. There are just nits regarding the clang 19 dependency:

Could you change the conditional dependency on clang 19 in the `extra-build-tools`  package to r52? (also see the follow up comment)

rzezeski:
Thanks for looking. Pushed the update.

hadfl (approved):

citrus-it (approved):

PR #1597, opened 2025-04-11
Bump atuin to version 18.5.0

papertigers:
This updates atuin to version 18.5.0 with some nice new features.

I removed the libc patch as the version in Cargo.lock is newer than what we were updating for before.
This built locally for me without issues minus the normal warnings about things like libpthread.so.


hadfl:
OmniOS ships rust 1.86, so this should be fine. We are in the middle of freezing for the upcoming r54 release. We'll get back to this PR once that's done.

oetiker (approved):

hadfl (approved):

PR #1604, opened 2025-05-07
Bump atuin to v18.6.0

papertigers:
This bumps atuin to the latest release aka [v18.6.0](https://forum.atuin.sh/t/release-v18-6-0/977).

I built this in my build VM without issue.

### DIFF
--- a/build/atuin/build.sh
+++ b/build/atuin/build.sh
@@ -17,7 +17,7 @@
 . ../../lib/build.sh
 
 PROG=atuin
-VER=18.5.0
+VER=18.6.0
 PKG=ooce/util/atuin
 SUMMARY="Magical shell history"
 DESC="Replaces your existing shell history with a SQLite database and "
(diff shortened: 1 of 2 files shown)


hadfl:
I assume you renamed the source file taken from GitHub? It seems atuin no longer prefixes version tags with a `v` so this needs to be removed in `download_source`.

papertigers:
You're right I used a `site.sh` file and I had pulled the files myself:
```
omnios@omnios-builder:~/extra-web/atuin$ ls -l
total 2652
-rw-r--r--   1 omnios   other     665686 Apr  9 11:15 v18.5.0.tar.gz
-rw-r--r--   1 omnios   other         81 Apr  9 11:15 v18.5.0.tar.gz.sha256
-rw-r--r--   1 omnios   other     667624 May  6 23:41 v18.6.0.tar.gz
-rw-r--r--   1 omnios   other         81 May  6 23:41 v18.6.0.tar.gz.sha256
```

If you look at the GitHub releases it seems like there is no standard (a sampling):
[18.6.0](https://github.com/atuinsh/atuin/releases/tag/18.6.0)
[v18.5.0](https://github.com/atuinsh/atuin/releases/tag/v18.5.0)
[18.5.0-beta.3](https://github.com/atuinsh/atuin/releases/tag/18.5.0-beta.3)

Do we want `build.sh` to just track the latest version scheme?

papertigers:
I am happy to drop the `v` in the build script but let me rephrase a different way.

How are you mirroring the tar.gz today? This seems like a manual step unless there's some sort of automation I am unaware of. This package isn't grabbing the source directly from GitHub.

citrus-it:
> How are you mirroring the tar.gz today? This seems like a manual step unless there's some sort of automation I am unaware of. This package isn't grabbing the source directly from GitHub.

Yes, it's manual (and to satisfy some licences that require us to provide the source if we're providing binaries). We just keep the names the same as the origin. It's going to be a pain if there is no consistency upstream, but let's switch to the format of the latest and see how it goes.

papertigers:
Sounds good to me, just pushed an update that drops the leading "v"

hadfl (approved):

PR #1628, opened 2025-07-12
Introduce centralised OOCEVER as distinct from RELVER

citrus-it:


citrus-it:
Discarded, we'll stick with `$RELVER` everywhere.

PR #1632, opened 2025-07-21
Helix: update to 25.07.01

snltd:
Bump to the new release of Helix. I built, installed, and tested on my own system, and it all looks good.


oetiker (approved):

oetiker (changes requested):
you have to update the packages.md too

PR #1642, opened 2025-09-14
Add micro text editor

extrowerk:
micro is a GO based terminal editor. The latest release is broken on Illumos, but the git master builds and works just fine. I'd like to get more experience with creating IPS packages, therefore I try to make an own recipe. 

The resipe is updated to point to a specific git commit instead of master. This makes the recipe reproducible.

Thanks to andyf for the help!


extrowerk:
Bumped to latest git commit.

PR #1652, opened 2025-10-07
Nginx: enable ngx_stream_ssl_module

extrowerk:
To SSL proxy IRC I'D like to enable the mentioned module.


citrus-it:
> NginX compiles, but at the packaging stage I get:

This is ok, just warnings about opportunities for clean up.

PR #1665, opened 2025-10-20
WIP: mg: add new recipe

extrowerk:
For some reason the configure file dissapears from the source tree. I definetely need somebody to explain me how the build system actually works here.


hadfl:
Just out of curiosity since this is yet another editor. Are you going to use the packages you are adding or is this more like an exercise? We usually have the policy that we require at least one consumer for a package to be added, to keep the burden of maintaining packages as small as possible.

PR #1696, opened 2025-12-20
pari: version 2.17.3

cstes:
This is a build script which I (David Stes) wrote for the PARI computer algebra system from the University of Bordeaux, France : see the URL on top of the build script.
 (https://pari.math.u-bordeaux.fr).

For OpenIndiana there exists a package, and for Debian Linux there exists a package that is maintained by Bill Alombert from the PARI group.

I am not a member of the PARI group but I have an interest in number theory and computer algebra, and also in OmniOS/Illumos.

The source code can be downloaded from the URL indicated in the build script, and I provide a patch for OpenIndiana which is not necessary I think on OmniOS which relates to the "out of source tree" build of the non-autoconf Configure script.

If patch 01 is applied then the package is broken (regeneration of the functions during the build does not correctly list functions, and the build succeeds but "gp" (the PARI command) is unusable.  So patch 01 is provided (for OpenIndiana) but not for OmniOS.

I tested on AMD Ryzen 7  (amd64 64bit) only and provide only a build script for amd64 currently.

The "testsuite" runs for me (without the optional data packages galois and other add on mathematical databases, these could be added later if there is any interest).  The optional databases can also be manually installed as explained in the PARI documentation.   For the testsuite there was a test that failed (env) but this seems to be solved by setting environment variables AAA and BBB which are tested by the env test from PARI.

There are a few errors during the build:   there is an error that -lpthread is not necessary any longer so I provide a patch 02-get_mt.patch which is applied to remove the unnecessary MT-LIBS=libpthread library in the case of --mt=pthread.  

There is also an error "CTF failed for opt/ooce/pari/bin/gp-2.17"

I could also provide patches for get°libc and get_readline because there are warnings or errors related to unnecessary -lm and -lncurses libraries (I am not sure whether readline requires libncurses on OmniOS but I suspect it does not, so the PARI get_readline could be patched).

I am interested in feedback (help) on improving the build.sh script and to make the package ported to other CPU's (32bit).

I wrote this text in English but I am  based in Belgium and I understand German or French, and I would even appreciate a reply in German for example , to practice my German.

Regards,
David


cstes:
I tested pari 2.17.3 and pari 2.8.0 alpha on OmniOS r151058 current version from 2026 and the packages compile with gcc OmniOS 151058 15.2.0 illumos.   The installation of OmniOS new version and PARI install (by compiling it following the install instructions of PARI) went fine.   The PARI package does not find GMP in the OmniOS location but the PARI configure script has switches to indicate the location of the GMP header file.

PR #1700, opened 2026-01-01
Syslog-NG: add 4.10.2 : WIP

extrowerk:
Syslog-NG builds with some minor patching.
TODO: The contrib folder have some Solaris 10 SMF XML files which we can probably reuse.


extrowerk:
This can be useful: https://lucamerello.wordpress.com/2014/06/13/solaris-10-how-to-build-and-install-syslog-ng/

PR #1713, opened 2026-02-13
xxhash: WIP

extrowerk:
The makefile checks for uname here: https://github.com/Cyan4973/xxHash/blob/dev/Makefile#L572
and then it tries to use ginstall, which is not available on my side. 
I belive there must be an easier way than patching it, but I have failed to find it.

#HELPNEEDED


extrowerk:
Unneeded

PR #1744, opened 2026-04-29
Soju: add v0.11.0

extrowerk:
A user-friendly IRC bouncer
https://soju.im/
https://codeberg.org/emersion/soju


extrowerk:
No idea why the check fails.

PR #1753, opened 2026-05-22
add-fish-build

snltd:
Add Fish 4.7.1.

I know there's already a Fish PR open, but it seems to have stalled, and this one is of the current version. I used the same original patch from OpenNevada, modified a little bit to get it working with 4.7.1.

Please check the licensing. There are four licenses in `doc_src/license.rst`, and I could not get "LGPL" or "PSFv2" to be recognised without modifying `doc/license`. (Removing the anchors for PSFv2; adding a new line for LGPL.) 

I chose to leave `doc/license` as it is, and specify "Various" license along with the two that were recognised. Obviously happy to change this if it's not suitable.

I've used this build as my daily driver on 151056 and 151058 with no issues.


citrus-it:
Thanks for putting this together.
I've pushed a small commit on top for you to look at which switches to pulling the source code from the omnios source mirror. That's something that helps with our legal people's peace of mind as many software licences require that we distribute source alongside binary packages.

I agree with what you've done around the licensing. It's a small risk of not picking up licence changes across updates but it's probably easier and safer than adjusting the patterns to pick up the formatting here.

If you're happy with this, please can you squash everything down into one commit and we'll get it in.

Since you use this, any work on the upstream nix crate to help with illumos support would also be appreciated so we can drop these patches longer term.

Thanks again for the PR.

snltd:
I had a look into the nix bug, and it turned out someone had already done the work, but it had never been tested, and fallen out of date. I tested the fix against fish, and fish seems build correctly.

I'm now running that fish build as my shell, and I've [opened a new nix PR](https://github.com/nix-rust/nix/pull/2799) that's fully up-to-date.

I'll have a look at putting together a patch for fish which uses my nix fork, as it's far cleaner than what we have here, but hopefully we'll get the proper fix merged into nix soon.

snltd:
I have [a fork of nix with working termios](https://github.com/snltd/nix/tree/fix/termios-from-bits-retain) in my Github. I've removed the revert-termios patch we originally had, and now we use the forked nix. It looks good to me.

Hopefully we'll get the termios fix merged into nix before too long, but I'll keep the fork around as long as it takes.

citrus-it (approved):
Thank you!

PR #1774, opened 2026-07-13
Nagios : update to 4.5.14

extrowerk:
~~4.5.13 have changed the directory name in the tarball, and i dont know how to adjust the build.sh for this case.~~

### DIFF
--- a/build/nagios/build.sh
+++ b/build/nagios/build.sh
@@ -17,7 +17,7 @@
 . ../../lib/build.sh
 
 PROG=nagios
-VER=4.5.0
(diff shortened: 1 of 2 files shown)


mightyspiky:
I understand there is an issue with 4.5.13 but is 4.5.12 gonna be available someday or is this a dead thread?

extrowerk:
I dont use Nagios, but wanted to contribute with this low hanging fruit. It needs a review but otherwise should be ok.

extrowerk:
Updated to the latest. It builds now just fine.

PR #1789, opened 2026-08-04
mariadb: pass MAP_NORESERVE when reserving the InnoDB buffer pool

citrus-it:


hadfl (approved):

andrewhotlab:
I spotted this problem to a friend which is close to the MariaDB developer team, and yesterday they filed this bug upstream: https://jira.mariadb.org/browse/MDEV-40921

PR #1812, opened 2026-09-07
Package glow

rosekanari:
Add [glow](https://github.com/charmbracelet/glow) 3.0.0, a terminal markdown reader from Charmbracelet.

Upstream has no Makefile (goreleaser/Taskfile), so `build()` calls `go build` directly with `-ldflags` for version and commit, then generates the man page and bash/zsh/fish completions from the binary (`glow man`, `glow completion <shell>`).

`CGO_ENABLED=0`, matching upstream's release binaries.

Built and tested on r151054, r151058 and bloody: `build.sh` rc=0, `pkglint` clean, `pkg verify` clean after install from a local repository; version string, rendering and completions exercised.

Package is `ooce/text/glow`; happy to move it to `ooce/util` if that fits better.


rosekanari:
Built and verified on r151054, r151058 and bloody; details folded into the description.

rosekanari:
Amended the commit to set `CGO_ENABLED=0`, as upstream does for all its release binaries. This also makes the binary identical across the three branches. Rendering a remote markdown file over HTTPS was checked on r151054 and bloody, so DNS resolution through the pure-Go resolver works.

PR #1820, opened 2026-09-13
Package fzf

rosekanari:
Add [fzf](https://github.com/junegunn/fzf)  0.74.4, a command-line fuzzy finder. No patches: upstream's Makefile already handles illumos.

Upstream's `install` target is a no-op, so the recipe installs the files itself: `bin/fzf` and `bin/fzf-tmux`, man pages, the bash completion under `share/bash-completion/completions/`, and the shell key bindings and zsh/fish integration under `share/fzf/shell/`, to be sourced by the user. Nushell files are skipped (nushell is not packaged). `tmux` is not made a dependency of `fzf-tmux`.

Built with `CGO_ENABLED=0`, matching upstream's release binaries: fzf does not use cgo, and this makes the binary identical across the three branches instead of varying with each branch's gcc.

Built and tested on r151054, r151058 and bloody: `build.sh` rc=0, `pkglint` clean, `pkg verify` clean after install from a local repository; `--filter`, `fzf-tmux` and the bash completion exercised.


rosekanari:
Note on the hash in the description: 83d7b398 is the build with Go 1.26.7. The three branches now ship Go 1.26.8, and the same recipe gives 0324dfd6 on all three — still bit-identical across r151054/r151058/r151059. The property claimed is cross-branch identity at equal Go version, not a fixed hash; the only bytes that move between 1.26.7 and 1.26.8 are the build IDs and the version string (0 bytes in .text, verified on one machine with both toolchains).

PR #1826, opened 2026-09-16
gitea 1.27.3

### LINKED ISSUE #1754: gitea: needs fix for CVE-2026-20750
There's an open CVE on gitea that currently requires an update to v1.26.2 (https://app.opencve.io/cve/CVE-2026-20750), but due to changes in Gitea, it's not currently possible to build 1.26 on Illumos (Vite/Rollup.js moved parts to Rust and currently disallow builds on "sunos" - does anyone know how to work with them to fix this?  As long as they aren't doing anything esoteric with sockets, Rust works fine here, doesn't it?)

I've requested the Gitea maintainers backport this bugfix to the 1.25 branch, and will submit a PR myself if needed.  Once a new v1.25 build is available, I'll update our package here.

rosekanari:
Update gitea from 1.25.5 to 1.27.3. Fixes omniosorg/omnios-extra#1754

Changes to the recipe:

- `download_source` instead of `clone_go_source`, fetching from the mirror's `gitea/` path. The upstream release archive (`gitea-src-<ver>.tar.gz`, the one gitea publishes for packagers) ships `public/assets` prebuilt. Since 1.26 the frontend is bundled with vite/rolldown, which does not build on illumos yet; the archive avoids that step. `set_nodever` dropped for the same reason.
- `sqlite_mattn` added to `TAGS`: 1.27 makes `modernc.org/sqlite` the default driver, and it is not published for illumos.
- `BUILD_DEPENDS_IPS+=` instead of `=`, so the Go dependency added by `set_gover` is not overwritten.

Built and tested (`build.sh` rc=0, `pkglint` clean, `pkg verify` clean) on r151054, r151058 and bloody. Also built as an unprivileged user outside a zone.Installed from a local IPS repository and exercised: user and repository creation, HTTP clone/push, issue, PR, merge, Actions workflow with `act_runner`.


rosekanari:
Also built on a fourth machine, and this one as an **unprivileged user** rather than as root in a zone: bare-metal OmniOS r151058q (Xeon E-2224, 4 cores), `build.sh` rc=0 in 3m34 on the same commit, Go 1.26.8, gcc 15.2.0. The three builds in the table above run as root inside a `build` zone with `ROOT_OK=1`; this one used the framework's normal non-root path, so the recipe is validated in both regimes.

Written with claude-code.

rosekanari:
The V8 issue numbers promised in the description, with one correction: the defect was **already tracked upstream** — I found the two issues before filing, both opened in June/July 2026 by V8 and emscripten people, so there is no new issue of ours to link.

- https://issues.chromium.org/issues/529880019 — *Cross-worker memory growth not observable with `--wasm-enforce-bounds-checks`*, which is the trapping form that breaks the wasm fallback here;
- https://issues.chromium.org/issues/533026477 — *`memory.size` should be an atomic operation maintaining sequential consistency*, the underlying stale-size problem.

Both are still open, and our reproduction still fails on an official canary `d8` (V8 15.6.3) that postdates the fixes landed on both; the measurements are attached there as comments. The illumos angle turns out to be useful to them rather than the other way round: illumos has no V8 trap-handler support, so the failing path is the default and the failure is deterministic (5 runs out of 5, in a VM and on bare metal), where both threads report reproductions that come and go.

For this PR nothing changes: the recipe builds the frontend from upstream's release archive and never enters that path.

build/gitea/build.sh
```diff
@@ -36,7 +44,9 @@ XFORM_ARGS="
     -DVERSION=$VER
 "
 
-BUILD_DEPENDS_IPS="
+# += rather than =: set_gover appends ooce/developer/go-1xx to this variable,
+# and a plain assignment here would discard it.
```
citrus-it: This comment isn't needed. Yes, the original = was a mistake, but chapter and verse and history don't belong in a comment here.

build/gitea/build.sh
```diff
@@ -27,7 +27,15 @@ PREFIX+=/$PROG
...
-set_nodever
+
+# No set_nodever: the upstream source release ships public/assets already built,
+# so make finds the frontend up to date and never invokes the node toolchain.
+# Building the frontend from source is not currently possible on illumos -- the
+# pull request for this change has the detail.
+
+# The source release, not a git clone: a clone has no prebuilt assets and would
+# force the frontend build.
+set_mirror https://github.com/go-gitea/gitea/releases/download
```
citrus-it: Please remove this. We'll provide the archive from the omnios mirror - I've seeded that for you.

build/gitea/build.sh
```diff
@@ -27,7 +27,15 @@ PREFIX+=/$PROG
...
 set_gover
-set_nodever
+
+# No set_nodever: the upstream source release ships public/assets already built,
+# so make finds the frontend up to date and never invokes the node toolchain.
+# Building the frontend from source is not currently possible on illumos -- the
+# pull request for this change has the detail.
+
+# The source release, not a git clone: a clone has no prebuilt assets and would
+# force the frontend build.
```
citrus-it: None of this comment is needed here, please remove.

build/gitea/build.sh
```diff
@@ -55,7 +65,14 @@ build() {
...
 
     logmsg "Building 64-bit"
-    TAGS="bindata sqlite sqlite_unlock_notify" logcmd $MAKE build \
+    # sqlite_mattn is required from 1.27 on. That release added a second
+    # driver, modernc.org/sqlite, and made it the default
+    # (models/db/driver_sqlite_modernc.go is //go:build !sqlite_mattn). It is
+    # pure Go with per-platform generated definitions, published for darwin,
+    # freebsd, linux, netbsd, openbsd and windows but not illumos, so the build
+    # fails with "undefined: TBitvec" and a few hundred siblings. The tag
+    # selects the cgo driver instead, which builds here.
```
citrus-it: Please drop this comment.

build/gitea/build.sh
```diff
@@ -83,7 +100,8 @@ install() {
 }
 
 init
-clone_go_source $PROG go-$PROG v$VER
+set_builddir $PROG-src-$VER
```
citrus-it: set_builddir should come near the top of this script - see other packages for examples.

citrus-it (changes requested):
Thank you for looking at updating this, we certainly need to be providing a later version.

In general, please avoid just pasting the entire output from Claude into a PR or getting it to open the PR on your behalf. While things like the testing matrices are appreciated, it's hard to read a wall of text like that and there is no evidence in the PR that I can see that a human was involved at all. A lot of the new comments are documenting history, or changes, which is typical of Claude but not generally helpful and often duplicates what the git history already tells us.

build/gitea/build.sh
rosekanari: Removed. Note that gitea/v1.27.3.tar.gz on the mirror is the tag archive, which has no prebuilt public/assets and so triggers the frontend build that fails on illumos. The recipe expects the packagers' archive from the release page, gitea-src-1.27.3.tar.gz (69,885,076 bytes, sha256 3283ae40dd1f7b09450bb5a56455e78106fe17f4211d254c7c0179b8927bf382), which ships the frontend already built. Could you put that one on the mirror instead, with its .sha256? With it in my local source cache the recipe builds cleanly on r151054, r151058 and bloody.

rosekanari:
Thanks for the review. 
Comments removed, set_mirror dropped, set_builddir moved up, description trimmed. 
One note on the mirror archive in the inline thread.

citrus-it (approved):

